PR #40, opened 2014-06-07
Py3

mpharrigan:
I had to make these changes to get pdbfixer to work with python 3. I also removed unnecessary imports from ui.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,11 @@
 language: c
+
+env:
+    matrix:
+        - CONDA_PY=2.7
+        - CONDA_PY=3.3
+        - CONDA_PY=3.4
+
 install:
     - sudo apt-get update -qq
     - sudo apt-get install -qq python-dev python-pip python-yaml g++ ftp
--- a/pdbfixer/__init__.py
+++ b/pdbfixer/__init__.py
@@ -1,1 +1,3 @@
-from pdbfixer import PDBFixer
+from __future__ import absolute_import
+from .pdbfixer import PDBFixer
+
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -28,6 +28,7 @@ DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
 OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
+from __future__ import absolute_import
 __author__ = "Peter Eastman"
 __version__ = "1.1"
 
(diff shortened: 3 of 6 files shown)


jchodera:
@peastman : Did you want to merge this before the `pdbfixer 1.1` release?  I think we're aiming for everything in omnia to be Python 3 compatible, aren't we?

mpharrigan:
iirc there are still some bytes / unicode issues with fetching and displaying the images in py3 that this pr does not address

kyleabeauchamp:
Hmm.  Given the py3k issues with yank, pymbar, and pdbfixer, we could always make omnia 1.0 a py27-only release and punt.

mpharrigan:
I'd still recommend merging this. It lets pdbfixer at least function, although it isn't as pretty

jchodera:
I think the py3 issues with yank, pdbfixer, and pymbar could be very
quickly addressed in a day.

kyleabeauchamp:
Let's discuss on pymbar tracker.  For me, the issue is that we don't need to be investing a day backporting py3k compatibility into pymbar 2.2 when we're getting ready for a new major release where we already plan to have py3k compatibility.  We have ~tens of tasks that could "just take a day", and each one delays the omnia release further.

jchodera:
Good point.

jchodera:
@peastman: How about merging this?

peastman:
Ok with me.  Is it ready to go?  The last I see in the discussion, there were still some problems that needed to be resolved.

jchodera:
Good point.  At the very least, python 3.3 and 3.4 should be added to the `.travis.yml` build matrix in `env`, like this:
https://github.com/msmbuilder/msmbuilder/blob/master/.travis.yml#L20-L24

Also, it looks like travis integration has either switched off or the travis results are so old that they were deleted.  Can you check this, @peastman?

mpharrigan:
It's confused because of the SimTk organization rename

jchodera:
I'm guessing @peastman is the one that needs to fix that.

jchodera:
@mpharrigan : Can you add this block to the `.travis.yml` in your PR?
```
env:
  matrix:
    - CONDA_PY=2.7
    - CONDA_PY=3.3
    - CONDA_PY=3.4
```
See, e.g.
https://github.com/msmbuilder/msmbuilder/blob/master/.travis.yml#L20-L24

peastman:
Where is it that the name needs to be changed?

mpharrigan:
I'll add that to the build matrix

I'm fixing the images as well

jchodera:
@peastman: Perhaps you can just check the webhooks that are configured for this project and make sure Travis-CI is enabled?

peastman:
It says it's enabled.

mpharrigan:
I'd also like to humbly suggest using an extant web framework for future projects (e.g. http://www.tornadoweb.org) instead of rolling our own :)

jchodera:
> I'd also like to humbly suggest using an extant web framework for future projects (e.g. http://www.tornadoweb.org) instead of rolling our own :)

PLEASE YES.

`tornado` is definitely on `conda`.

jchodera:
All things pass on all Pythons!  Yay!

mpharrigan:
This should be ready to merge

PR #72, opened 2015-02-11
Test commit

rmcgibbo:


jchodera:
That seems to have reawakened travis!

rmcgibbo:
Perhaps the Travis-CI is confused because #40 was originally a PR to the repo when it was under a different name.

rmcgibbo:
Perhaps easiest to close #40 and reopen a new PR from the same git branch.

mpharrigan:
It'll probably just work with a new commit to that pr

PR #73, opened 2015-02-18
Eliminated most dependence on PdbStructure

peastman:
This simplifies the code, and also will make it easier to add support for loading PDBx files.  It requires the changes to Topology I just merged into OpenMM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,11 @@
 language: c
+
+env:
+    matrix:
+        - CONDA_PY=2.7
+        - CONDA_PY=3.3
+        - CONDA_PY=3.4
+
 install:
     - sudo apt-get update -qq
     - sudo apt-get install -qq python-dev python-pip python-yaml g++ ftp
--- a/pdbfixer/__init__.py
+++ b/pdbfixer/__init__.py
@@ -1,1 +1,3 @@
-from pdbfixer import PDBFixer
+from __future__ import absolute_import
+from .pdbfixer import PDBFixer
+
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -6,7 +6,7 @@ Simbios, the NIH National Center for Physics-Based Simulation of
 Biological Structures at Stanford, funded under the NIH Roadmap for
 Medical Research, grant U54 GM072970. See https://simtk.org.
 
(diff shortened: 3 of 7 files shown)


peastman:
The build is failing because it isn't getting the latest changes to OpenMM.  Do I need to do something to fix that?

kyleabeauchamp:
AFAIK, I don't think we have been building conda packages for every openmm commit--at least not successfully.  (https://binstar.org/omnia/openmm-dev  https://binstar.org/omnia/openmm)

That's something we could do if necessary, though.

kyleabeauchamp:
Personally, I'm happy if you can confirm that tests run locally.

rmcgibbo:
Travis is grabbing the conda package called `openmm`, which is the 6.2 release. https://github.com/pandegroup/pdbfixer/blob/master/.travis.yml#L15

peastman:
I recall there being a lot of discussion about this issue for Omnia: how to handle development versions of projects that depend on development versions of other projects.  I don't remember what was decided though.

kyleabeauchamp:
I think John has the jenkins infrastructure in place to do this.  AFAIK there are tests being run on every commit, it's just the packages aren't uploaded.  

Regarding "development versions"--you are correct, there's also the question of whether you want to run Travis against released packages or development packages.  IMHO, we are still using a temporary workaround (`xyz-dev`).  Whenever possible, IMHO I prefer to test against releases, but this is one of those cases where that's not feasible.

rmcgibbo:
For some of the packages, the travis build, if it succeeds, pushes the package binary to as `<name_of_package>-dev` to binstar, but it doesn't look like OpenMM does this (successfully).

jchodera:
Our current scheme is to use travis to push `<packagename>-dev` versions to the `omnia` binstar channel.  Downstream packages that use `<packagename>` then choose whether they have travis test against `<packagename>` (the latest stable release) or `<packagename>-dev` (the cutting-edge development version) on a package-by-package basis.

We should definitely fix the OpenMM travis to get `openmm-dev` pushed correctly on commits if that isn't working right now.

peastman:
Looks like it will take a bit longer to get the conda issues resolved, so I'll just go ahead and merge this.  I ran the tests on my own computer and they all passed.

Issue #75, opened 2015-02-24
Have travis build pdbfixer-dev package and push to binstar


jchodera:
We have other projects build `*-dev` packages and push them to binstar so we can test development versions easily.  Would be great to do this here too.

peastman:
Sounds like a good idea.  I don't know what's involved in that, but feel free to send a PR.

jchodera:
I'm happy to do this, but on the condition that you clarify up front what behavior is and is not acceptable, since just ended in a lot of wasted effort on my behalf with openmm.

There are a few different schemes in use by omnia projects at the moment.

Yank, for example, first builds the conda package using `devtools/conda-recipe`:
https://github.com/choderalab/yank/blob/master/.travis.yml

This recipe runs the tests, failing if they don't pass:
https://github.com/choderalab/yank/blob/master/devtools/conda-recipe/meta.yaml#L43-L51

If all is well, the `devtools/ci/after_success.sh` script uploads the conda package to binstar:
https://github.com/choderalab/yank/blob/master/devtools/ci/after_sucess.sh

Would this behavior be acceptable?

peastman:
The main requirement is that it not interfere with testing.  That means, for example:

* A problem with conda or binstar shouldn't prevent the tests from getting run, or cause them to fail inappropriately.

* It shouldn't cause travis to take much longer.  (This isn't likely to be an issue here, since "building" PDBFixer takes seconds at most.  Also, changes to PDBFixer are much less frequent than changes to OpenMM, so it's less important if it adds a minute or two to the process.)

jchodera:
> A problem with conda or binstar shouldn't prevent the tests from getting run, or cause them to fail inappropriately.

In the build strategy I proposed above, an error in the conda recipe `meta.yaml` specified prerequisites or the `build.sh` script (which should just contain `python setup.py install`) would technically cause a failure and the tests not to be run.  So, formally, someone could mess up these files or add an unspecified prerequisite and the build will fail and the tests not to be run, but it seems very unlikely that this would happen once we got it working in the first place.

Now, does this mean that this strategy would be rejected here, or is it OK for this case?

jchodera:
Working on this, assuming that this strategy will work for you.

peastman:
Ok.  If someone makes a change to the repository and that breaks something, that's not a problem.  That just means there's something wrong with their changes.

peastman:
Can this be closed now?

jchodera:
This is working!

PR #77, opened 2015-02-25
Have travis build pdbfixer-dev conda package and push to binstar

jchodera:
See #75.

Still testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,33 @@
 language: c
 
 env:
-    matrix:
-        - CONDA_PY=2.7
-        - CONDA_PY=3.3
-        - CONDA_PY=3.4
+  matrix:
+    - CONDA_PY=2.7
+    - CONDA_PY=3.3
+    - CONDA_PY=3.4
 
 install:
-    - sudo apt-get update -qq
-    - sudo apt-get install -qq python-dev python-pip python-yaml g++ ftp
-    - sudo pip install conda eventlet
-    - sudo conda init
-    - conda config --add channels http://conda.binstar.org/omnia
-    - conda create --yes --name test nose setuptools numpy biopython pyflakes openmm-dev
-    - export PATH="$HOME/envs/test/bin/:$PATH"
-    
+  - source devtools/ci/install.sh
+  - export PYTHONUNBUFFERED=true
+#    - sudo apt-get update -qq
+#    - sudo apt-get install -qq python-dev python-pip python-yaml g++ ftp
+#    - sudo pip install conda eventlet
+#    - sudo conda init
+#    - conda config --add channels http://conda.binstar.org/omnia
+#    - conda create --yes --name test nose setuptools numpy biopython pyflakes openmm-dev
+#    - export PATH="$HOME/envs/test/bin/:$PATH"
+
 script:
-    - python setup.py install
-    - cd tests
-    - nosetests --with-doctest
+  # this builds the binary, unpacks it, and runs the tests
+  - conda build devtools/conda-recipe
+  # - python setup.py install
+  # - cd tests
+  # - nosetests --with-doctest
 
 after_script:
-    - pyflakes pdbfixer/*.py
+  - pyflakes pdbfixer/*.py
+
+after_success:
+  - echo "after_success"
+  - bash -x devtools/ci/after_sucess.sh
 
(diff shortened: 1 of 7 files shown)


jchodera:
Travis is showing a bunch of nosetest failures.  I can reproduce this on my machine (using the `openmm-dev` conda package).

@peastman : Do the nosetests pass on your machine?  Try this:
```
nosetests pdbfixer --nocapture --verbosity=2 --with-doctest
```

jchodera:
It looks like the current master has a lot of broken nosetests, which are causing the failure.

@peastman: How do you want to handle this?  Do you want to fix the nosetests in a separate PR?

peastman:
It looks like it's building against an out of date version of OpenMM:

    AttributeError: 'Chain' object has no attribute 'id'

jchodera:
So nosetests pass on your machine? Could be that openmm-dev is outdated.

kyleabeauchamp:
The timestamps on openmm-dev packages are different for OSX and linux, not sure if this is an issue or not

jchodera:
> The timestamps on openmm-dev packages are different for OSX and linux, not sure if this is an issue or not

They're built by different jenkins scripts.  But should be +- 1 day of each other.

peastman:
The PR that added the "id" fields was merged into OpenMM over a week ago.

jchodera:
The builds were 5 days ago (`osx`) and 19 hours ago (`linux`) but there were some old files on binstar as well and it is possible that binstar got confused as to which ones to serve up:
https://binstar.org/omnia/openmm-dev/files

I've removed the older files.  Doing more testing.

jchodera:
@peastman : I think we're down to actual errors now.  Can you take a look at the travis output?

peastman:
It looks like that's been broken for ages!  I didn't see it when I ran the tests locally, because it wasn't running the doctests.

@kyleabeauchamp created the problem on May 21 last year (https://github.com/pandegroup/pdbfixer/commit/4a38eb7d9c73c6d709a3d8ce70d29f41b9b9a367) when he added the mandatory chain_id argument, but didn't update the doctest accordingly.  So why hasn't this been failing ever since?  I'm not sure what's going on with that.

peastman:
It also looks like there are some Python 3 issues.  The Python 2.7 build only has that one failure, but the 3.3 build lists a bunch of others.

kyleabeauchamp:
hmm, will take a look. 

Actually, it might make sense to refactor some of the changes that I implemented to use the latest features from openmm

jchodera:
Perhaps doc tests were not being run?

If you want to fix in a separate PR first, let me know.

peastman:
The --with-doctest flag has been included since March of last year, so they should have been.  Let's try to resolve the problems first, then merge this when we get a clean build.

jchodera:
> Let's try to resolve the problems first, then merge this when we get a clean build.

Just to be clear, you are saying you want to resolve the doctest failures in a separate PR first, then merge #77 after you have tests pass on a local machine---is that right?

peastman:
Correct.  Kyle says he's looking into the doctest problem right now.  As for the Python 3 changes, I think they may be related to the recent PR from @mpharrigan.

kyleabeauchamp:
see #78

jchodera:
OK, looks like Python 2.7 passes now!  

I propose we fix the 3.x issues in a separate PR.

jchodera:
We can merge this now.

jchodera:
@peastman Apologies to be a pain, but can we merge this?  Would be good to get this package built so we can use it!

peastman:
The doctest is now fixed, but it's still failing on Python 3.  The PR Kyle just sent did *not* have that problem, so I think it must somehow be caused by your changes.  Also, it seems to have started filling up the log by printing out XML files.  I don't think it did that before.

kyleabeauchamp:
It looks like my PR was not running all of the tests?

```
Ran 11 tests in 9.731s
```

```
Ran 8 tests in 309.949s
```

jchodera:
Ah, got it. Will investigate.

peastman:
It looks to me like travis has *not* actually been running the doctests.  I'm not sure why: it's been specifying the --with-doctest flag.  But nonetheless it appears not to have been running them.  This PR adds a setup.cfg file, and that has caused it to start running them.  And that is where the Python 3 errors are happening.

jchodera:
Can we fix the py3 issues in a separate PR? I know very little about py3...

peastman:
Sure.  It looks like that error isn't actually caused by this PR, just made visible by it.

PR #80, opened 2015-02-26
Changes required to upload to binstar

jchodera:
Needed to make some more changes to authorize binstar upload.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
 
   global:
     # BINSTAR_TOKEN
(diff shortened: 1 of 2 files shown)


jchodera:
The failure here seems to have simply been due to a network error:
```
    URLError: <urlopen error [Errno 110] Connection timed out>
```

peastman:
Looks like the Python 2.7 build failed due to a networking error.  I'll go ahead and merge anyway, since your changes look pretty safe.

PR #83, opened 2015-03-02
Simplified examples

peastman:


peastman:
Alright, why is this not working?  Does it have something to do with the conda build?  Is it running the doctests against the conda installation (which doesn't have the test.pdb file) rather than the repository (which does)?

peastman:
Also, why is it printing the full content of soft.xml to the log?

kyleabeauchamp:
1.  Yes.  We've been installing things inside the python egg.  See, for example, https://github.com/mdtraj/mdtraj/blob/master/mdtraj/tests/test_gro.py#L40

kyleabeauchamp:
Regarding the XML: I have no idea why that keeps getting printed

peastman:
Is there any way to make this work without actually installing the test file as part of the application?

jchodera:
Can you just get by with using [`urlretrieve`](https://docs.python.org/2/library/urllib.html#urllib.urlretrieve) for now?
```python
import urllib
urllib.urlretrieve ("http://www.example.com/songs/mp3.mp3", "mp3.mp3")
```

Otherwise, I'd say we should be distributing data and test files along with the conda package so that we can use it in nosetests as well.

jchodera:
If you want to go the route of distributing data and test files in the package, let us know and we can help set up that structure.

kyleabeauchamp:
As a nosetest, one could skip the test if the PDB file is missing.  For a doctest, I'm not sure if it's as easy.

IMHO, it's not a bad idea to include the necessary test file as part of the application, as it allows us to test conda packages during package building and potentially after installation.

jchodera:
> IMHO, it's not a bad idea to include the necessary test file as part of the application, as it allows us to test conda packages during package building and potentially after installation.

I completely agree with this.

peastman:
> IMHO, it's not a bad idea to include the necessary test file as part of the application

Can we at least put it in the tests directory then?  What do I need to change to make that work?

jchodera:
Sorry for not being able to help earlier. We had a visitor here today. Will
get on this later tonight!

Question: Can we move the tests into pdbfixer with the layout

pdbfixer/tests - nosetests
pdbfixer/tests/data - test datafiles

For other project, this means we can run the nosetests after conda
installation, which is useful for debugging!

jchodera:
See #84

PR #84, opened 2015-03-03
Reorganize tests; simplify doctests.

jchodera:
Trying to fix #83


jchodera:
How's this?

kyleabeauchamp:
The py3k bugs are twofold.  One is the known buffering error.  The other is related to a broken nose import:

```
ImportError: cannot import name assert_items_equal
```

kyleabeauchamp:
I guess we're not fixing py3k right now then?  If so, this looks reasonable to me.  We can refine it more as we fix the other things

jchodera:
I think we're fixing the py3 issues in a separate PR.  The py2 tests pass.

peastman:
What changed?  Those tests weren't failing before.

kyleabeauchamp:
Are you sure?  The travis CI badge on the `master` branch claims that the py3K builds are failing:

https://travis-ci.org/pandegroup/pdbfixer

kyleabeauchamp:
Actually, it's hard to tell what build corresponds to master, but here's a recent PR:

https://travis-ci.org/pandegroup/pdbfixer/builds/52349632

peastman:
Exactly.  Look at the failures in that PR.  They were all in the doctests.  Those are precisely the failures I sent my most recent PR to fix.  The new errors are in unit tests, not doctests.

jchodera:
Here's one issue:
http://stackoverflow.com/questions/7473071/unittests-assertequal-and-iterables-only-check-the-contents
```
In python 3 assertItemsEqual method is named assertCountEqual.
```

peastman:
This is the line it's actually failing on:

    from nose.tools import ok_, eq_, raises, assert_items_equal

It's failing to import the nose function assert_items_equal(), *not* the unittests function assertItemsEqual().

jchodera:
I'm trying to fix this now.

jchodera:
In any case, I have no idea how my changes broke this for python 3.  I can't see how that could happen.

jchodera:
Looks like `assert_items_equal` was not present in the py3 `nose.tools`, but `assert_list_equal` is.

peastman:
Looks like there's a second use of assert_items_equal() that you missed.

jchodera:
Sorry, tests passed locally but had forgotten to commit the second fix.

peastman:
I'm inclined to merge this, then fix the one remaining problem in a separate PR.  Everyone ok with that?

jchodera:
Please do---I have no idea how to fix that.

peastman:
I think that last error is just because the default argument for NamedTemporaryFile() is mode='w+b'.  If we explicitly specify it and don't include 'b', that ought to fix it.

Issue #54, opened 2014-08-10
Add option to retain residue numbering and chain ids; should be default


jchodera:
Currently, pdbfixer spits out PDB files renumbered to start from 1 and chain id A.  This is probably not what we want, since pdbfixer is just supposed to make it possible to simulate the system while retaining as much information as the user desires, rather than changing information like residue numbers and chain ids.

jchodera:
Did @peastman already make changes to have PDBFixer retain residue numbering?  I'm noting issues with this...

peastman:
No, I haven't done this.

jchodera:
What would it take to do this?

jchodera:
Any thoughts here?  We could really use this.

peastman:
This is more challenging that it seems at first.  Somehow we need to guarantee that we're outputting valid, consistent identifiers.  Atom and residue ids need to be numeric, positive, in order, no more than 4 (for residues) or 5 (for atoms) digits, and unique.  Chain ids need to be single upper case letters, and again must be unique.  Usually the initial ids you load from the input file will satisfy that, but as soon as you start doing any editing of the topology it becomes a lot harder.  And even the input ids won't necessary be valid.  For example, if you load a PDBx file, the chain ids will be arbitrary strings.  (I know, I haven't implemented PDBx input yet, but it's one of the next things on my list.)

Right now the way it handles this is for the PDB writer to just ignore existing ids and generate new, guaranteed valid ones while writing the output file.

jchodera:
> Right now the way it handles this is for the PDB writer to just ignore existing ids and generate new, guaranteed valid ones while writing the output file.

This comes at the price of a huge amount of information loss and potentially making the output unable to be analyzed in a useful way.

I agree that there are challenges, but it seems like adding a mode that does the following would be at least conceptually straightforward:
* Renumber all atoms consecutively
* For inserted residues, name the residues by their corresponding three-letter-code from the `SEQRES` section, number them sequentially from the previous residue (throwing an Exception if this causes something weird to happen), and use the same chain identifier as the previous residue

[...]

PR #86, opened 2015-03-04
Preserve residue and chain IDs

peastman:
This fixes https://github.com/pandegroup/pdbfixer/issues/54#issuecomment-77052328.


peastman:
When using the GUI and command line interface, it always preserves IDs.  When using the scripting interface, it creates a Topology with the correct IDs, but you then need to pass keepIds=True to PDBFile.writeFile().

jchodera:
Awesome, thanks!

Issue #95, opened 2015-07-18
Add command-line argument for specifying pdbid instead of PDB filename


jchodera:
This could be useful and easy to add.

jchodera:
This has been added.

PR #96, opened 2015-07-18
Added --pdbid, --url, and --verbose flags.

jchodera:
* Users can now specify '--pdbid pdbid' flag in command-line to retrieve PDB file from RCSB
* Users can also specify '--url URL' to retrieve from custom URL.
* --vebose flag for command-line now prints out some information about what is going on; maybe we can convert this to logging later

### DIFF
--- a/Manual.html
+++ b/Manual.html
@@ -98,6 +98,8 @@ When run with no arguments, it launches the user interface.  If any arguments ar
 
 Options:
   -h, --help            show this help message and exit
+  --pdbid=PDBID         PDB id to retrieve from RCSB [default: None]
+  --url=URL             URL to retrieve PDB from [default: None]
   --output=FILENAME     output pdb file [default: output.pdb]
   --add-atoms=ATOMS     which missing atoms to add: all, heavy, hydrogen, or
                         none [default: all]
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/pandegroup/pdbfixer.svg)](https://travis-ci.org/pandegroup/pdbfixer)
+[![Build Status](https://travis-ci.org/pandegroup/pdbfixer.svg?branch=master)](https://travis-ci.org/pandegroup/pdbfixer)
 
 PDBFixer
 ========
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -2,6 +2,10 @@ package:
   name: pdbfixer-dev
   version: dev
 
+build:
+  entry_points:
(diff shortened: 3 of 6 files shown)


jchodera:
Implements #95

jchodera:
Hm.  There don't seem to be any nosetests of the CLI yet.

jchodera:
I've added some basic nosetests.

jchodera:
These test failures appear to be stochastic and related to sporadic failures to connect to the RCSB.  Perhaps having some form of memoization or caching of PDB files locally could be a way to easily avoid this issue.

jchodera:
This should be good to merge.

peastman:
Looks good.  I'm rerunning the Python 3.3 tests now.

peastman:
I reran them twice, but got errors both times.  I wonder if there's a way we could make this more robust against that sort of failure?

rmcgibbo:
in mdtraj, when we load from a url, i recall that we use a wrapper that
tries up to 3 times or so to grab the url before throwing

peastman:
I don't think all of the errors are caused by network problems.  This, at least, looks like a genuine bug in the code:

```
  File "/home/travis/miniconda/envs/_test/lib/python3.3/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/_test/lib/python3.3/site-packages/pdbfixer/tests/test_cli.py", line 54, in test_url
    run_cli('--url "http://www.rcsb.org/pdb/download/downloadFile.do?fileFormat=pdb&compression=NO&structureId=1LE1"')
  File "/home/travis/miniconda/envs/_test/lib/python3.3/site-packages/pdbfixer/tests/test_cli.py", line 31, in run_cli
    message += e.output
TypeError: Can't convert 'bytes' object to str implicitly
```

jchodera:
Good catch, @peastman!

peastman:
Looking at the errors, I notice that test_build_and_simulate has no problems, even though it downloads tons of files.  Most of the problems happen in test_remove_chains.  That one only uses one file... and downloads it ten times!  I wonder if the server is detecting lots of requests for the same URL from the same computer and blocking them?  This could be a DDOS protection mechanism.

Perhaps we should modify that test case to only download the file once and save it to a temporary file.

peastman:
I'll go ahead and merge this, then send a separate PR for the test case.

jchodera:
> Perhaps we should modify that test case to only download the file once and save it to a temporary file.

I think it would be more useful to have `PDBFixer` use memoization or caching, since this would prevent us from having to implement this mechanism in each test one might want to apply to the same PDB.

PR #97, opened 2015-07-20
Test case doesn't download PDB as many times

peastman:
This will hopefully help to address the timeouts we were getting in #96.


peastman:
Well, it's now only getting one timeout instead of several...  Let's see what else I can do.

peastman:
Everything passed!

PR #98, opened 2015-07-21
Migrate to the new container architecture.

swails:
I did this for ParmEd awhile back, and @rmcgibbo then did it for MDTraj.  They suggest doing this, and it *does* cut build times roughly in half (*and* builds start in seconds now).

I think that under the hood, the difference between the new container and legacy architectures in Travis is roughly (exactly?) the difference between Docker and Vagrant, respectively.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: c
 
+sudo: false
+
 env:
   matrix:
     - CONDA_PY=2.7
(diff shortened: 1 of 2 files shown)


swails:
Cut build times here in half, too.  `sudo` wasn't actually used for anything here, so migrating was trivial.

rmcgibbo:
+1

peastman:
Very cool!

PR #103, opened 2015-08-04
Fix issues with travis-ci and binstar upload of dev package.

jchodera:
This should hopefully fix issues with `pdbfixer-dev` not being pushed to binstar.


jchodera:
Seems to have worked:
https://binstar.org/omnia/pdbfixer-dev/files

PR #104, opened 2015-08-04
Use openmm >=6.3.0 in pdbfixer-dev package, not openmm-dev

jchodera:
We had previously been using `openmm-dev` for the `pdbfixer-dev` conda package.  Now we use `openmm >=6.3.0`.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -10,12 +10,12 @@ requirements:
   build:
     - python
     - setuptools
-    - openmm-dev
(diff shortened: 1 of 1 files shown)


peastman:
Why would we want to do this?  The latest development code for PDBFixer often depends on the latest development code for OpenMM.

jchodera:
> The latest development code for PDBFixer often depends on the latest development code for OpenMM.

OK, we can keep this then!

I was just surprised when it overwrote the openmm 6.3 I had installed, but it makes sense to keep it this way.

PR #107, opened 2015-08-06
[WIP] Added support for loading PDBx/mmCIF files

peastman:
This is functional and mostly complete, but I have some questions about the implementation.

First, the way it distinguishes between PDB and PDBx files.  The way I've written it right now, the only API change is when you're passing in a File object, in which case there are separate "pdbfile" and "pdbxfile" arguments.  If you specify either a filename or URL, it looks for filename extensions to decide what format it is.  Does that seem reasonable?  If so, are there additional extensions it should look for?

Alternatives would be 1) to have new arguments/UI/command-line-options for specifying the file type, or 2) to try to determine it based on the file contents.  The first one seems rather awkward to me, and the second one could be hard to do reliably.

Second, if the user specifies a PDB ID, should it download the PDB or PDBx version?  Both are available, and there are often differences between them.  Is there any general rule for which one tends to be "more accurate"?  Another consideration that our PDBx parsing is still pretty new and therefore more likely to contain bugs.  On the other hand, very large systems can't be specified with a single PDB file, so they get split into multiple files, and we don't handle that.  PDBx doesn't have that problem.


swails:
> Does that seem reasonable?

Seems reasonable to me.  ParmEd looks at the first couple lines of the file and determines the file type from that.  It's worked for every PDB and PDBx/mmCIF file I've thrown at it so far.  Code for PDB file [here](https://github.com/ParmEd/ParmEd/blob/master/parmed/formats/pdb.py#L66-L94) and for PDBx/mmCIF file [here](https://github.com/ParmEd/ParmEd/blob/master/parmed/formats/pdb.py#L753-L777)

There are ways to break it, but it'll work for any and every PDBx/mmCIF or PDB file you feed it from RCSB or wwPDB.

> Second, if the user specifies a PDB ID, should it download the PDB or PDBx version?

I tried a handful of PDB IDs:

```Python
import parmed as pmd
import numpy as np

np.abs(pmd.download_PDB('PDBID').get_coordinates() - 
       pmd.download_CIF('PDBID').get_coordinates()).max()

# Always returns 0.0
```

So the atoms are in the same location and same order in both files.

> Both are available, and there are often differences between them. Is there any general rule for which one tends to be "more accurate"? Another consideration that our PDBx parsing is still pretty new and therefore more likely to contain bugs. On the other hand, very large systems can't be specified with a single PDB file, so they get split into multiple files, and we don't handle that. PDBx doesn't have that problem.

Is this still true?  I thought that large systems were no longer split (as of like December last year, based on conversations/talks I've had with PDB scientists at Rutgers) and were only available as mmCIF.

There's also a push to go to mmCIF, so if we're downloading from the RCSB, I don't see a good reason *not* to go with the RCSB CIF file using the RCSB CIF reader.  Just my 2c of course.  And since the main parser for PDBx/mmCIF files was written by John Westbrook from the RCSB quite some time ago, I suspect it's more battle-tested than expected based on how short a time it's been in OpenMM.

peastman:
Perhaps I should have it first look for filename extensions, then if it doesn't find one it could look at the file content to try to guess.  That way even if the guessing isn't completely robust, it will only make things more robust than if we rely on filenames alone.

>  And since the main parser for PDBx/mmCIF files was written by John Westbrook from the RCSB quite some time ago, I suspect it's more battle-tested than expected based on how short a time it's been in OpenMM.

His code is probably very robust.  But I have a lot less confidence in my code for processing the file data and extracting relevant information!

Long term, we should definitely move to PDBx.  But in the short term, I'm worried about introducing bugs into a feature that used to work reliably.  So I suggest initially downloading PDB, with the plan of switching to PDBx once the relevant OpenMM/PDBFixer code is more mature.

peastman:
> So the atoms are in the same location and same order in both files.

In my tests the atom locations do match, but other aspects of the files can differ.  For example, chain identifiers, or the grouping of HETATM residues into chains, or the ordering of chains.

swails:
> Long term, we should definitely move to PDBx. But in the short term, I'm worried about introducing bugs into a feature that used to work reliably. So I suggest initially downloading PDB, with the plan of switching to PDBx once the relevant OpenMM/PDBFixer code is more mature.

+1.  Perhaps add an optional argument to go after the CIF file instead.  That way, we can battle-test the CIF parser more easily and when the transition *does* take place, it'll be easier.  Also, some files are *only* available in CIF, so it would be nice to be able to download those structures as well.

peastman:
Specifying PDB ID as XXXX is really nothing more than a shortcut for giving it the URL http://www.rcsb.org/pdb/files/XXXX.pdb.  And if you want the PDBx version, specify that same URL but change the filename extension to .cif.

swails:
> And if you want the PDBx version, specify that same URL but change the filename extension to .cif.

Fair enough.

peastman:
Ok, it now looks at file content if the filename isn't clear.

Issue #109, opened 2015-08-28
Residue mutations fail when PDB files contain repeated chain entries


jchodera:
The following script
```
from pdbfixer import PDBFixer
from simtk.openmm.app import PDBFile
fixer = PDBFixer(pdbid='1OL5')
fixer.findMissingResidues()
fixer.findMissingAtoms()
fixer.addMissingAtoms()
fixer.applyMutations(['TPO-287-THR','TPO-288-THR'], 'A')
```
fails with the error
```
Traceback (most recent call last):
  File "pdbfix1OL5.py", line 10, in <module>
    fixer.applyMutations(['TPO-287-THR','TPO-288-THR'], 'A')
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/pdbfixer-1.2-py2.7.egg/pdbfixer/pdbfixer.py", line 756, in applyMutations
    index = resSeq_to_index[resSeq]
KeyError: 287
```
This appears to be due to the fact that the PDB file contains chain A (protein), chain B (another protein), and then another chain A (solvent) and chain B (solvent).  This causes the `resSeq_to_index` dict to be populated with only the *last* chain `A` (solvent) rather than all residues belonging to chain `A`:
```
[...]

PR #110, opened 2015-08-28
Fix issue with mutations when chain appears in multiple places in PDB file

jchodera:
This addresses #109

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -775,9 +775,8 @@ class PDBFixer(object):
             deleteAtoms = [] # list of atoms to delete
 
             # Find atoms that should be deleted.
-            for residue, replaceWith in residue_map.iteritems():
-                if residue.chain.index != chain_number:
-                    continue  # Only modify specified chain
+            for residue in residue_map.keys():
+                replaceWith = residue_map[residue]
                 residue.name = replaceWith
                 template = self.templates[replaceWith]
                 standardAtoms = set(atom.name for atom in template.topology.atoms())
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -742,13 +742,15 @@ class PDBFixer(object):
         index_to_old_name = dict((r.index, r.name) for r in self.topology.residues())
         index_to_new_residues = {}
 
+        # Retrieve all residues that match the specified chain_id.
         # NOTE: Multiple chains may have the same chainid, but must have unique resSeq entries.
         resSeq_to_residue = dict() # resSeq_to_residue[resid] is the residue in the requested chain corresponding to residue identifier 'resid'
-        for (chain_number, chain) in enumerate(self.topology.chains()):
+        for chain in self.topology.chains():
             if chain.id == chain_id:
-                for (residue_number, residue) in enumerate(chain.residues()):
+                for residue in chain.residues():
                     resSeq_to_residue[int(residue.id)] = residue
 
+        # Make a map of residues to mutate based on requested mutation list.
         residue_map = dict() # residue_map[residue] is the name of the new residue to mutate to, if a mutation is desired
         for mut_str in mutations:
             old_name, resSeq, new_name = mut_str.split("-")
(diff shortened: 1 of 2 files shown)


jchodera:
This can be merged now.

jchodera:
Can we merge this, @peastman?

peastman:
Sorry, I'm still clearing out my email backlog from the weekend!  Could you explain what's going on here?  What is it supposed to mean when you have two different chains with the same ID?  Isn't that illegal?

swails:
> What is it supposed to mean when you have two different chains with the same ID?

I'm not entirely sure.  There are two possibilities:

1. A *change* in chain is all that matters, so the fact that the chain sequence goes A -> B -> A -> B could mean that there are 4 chains, and a new chain is simply different than the last one.
2. The second A in the above example is the solvent corresponding (in some way?) to chain A, while the second B is the solvent corresponding (in some way?) to chain B.

peastman:
I looked at this file, and there are clearly four chains, not two.  There's a TER record at the end of the first chain A, and another at the end of the first chain B.  So it isn't "going back" to the same chain and adding more residues to it.  The chain has already been marked as finished.

More generally, I think option 1 is safest.  Whenever the chain ID changes, we should interpret that as the start of a new chain.

jchodera:
As noted in #109, the solvent is specified in the second set of A and B chains in the case of 1OL5.  Clearly, the intention was to have two chains, since the residue numbers for the solvent do not overlap with the residue numbers specified in the first set of A and B chains.

You can interpret a change in ID as the start of a new chain, but you still need to keep the chain identifiers in the PDB file so that users can select out the chains they want using the IDs specified in the PDB.

jchodera:
In any case, how you choose to deal with this at the `PDBFile` level does not impact this PR, which will function fine regardless of whether you decide to read these in as two chains or four chains.

jchodera:
To clarify: The intention was clearly to name the waters the same as the polypeptide chains.

The [PDB format spec](http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#TER) says that the polypeptide chains must be terminated with `TER` records, but it doesn't say that more records of other molecules but the same `chainid` cannot follow them.

I've asked the RCSB for clarification of whether this is the desired behavior or not, but I think we can still go ahead and merge this.  All I did was change the mutation function to behave in the manner expected.

peastman:
I'm trying to understand your code, but either I'm missing something, or there's something wrong with it.

```Python
        chain_numbers = list()
        resSeq_to_index = dict()
        for (chain_number, chain) in enumerate(self.topology.chains()):
            if chain.id == chain_id:
                chain_numbers.append(chain_id)
                for (residue_number, residue) in enumerate(chain.residues()):
                    resSeq_to_index[int(residue.id)] = residue_number
```

First, what is the `chain_numbers` list supposed to be for?  You create it and add values to it, but then never use it for anything.  And (assuming it really is supposed to be used for something), shouldn't you add `chain_number`, not `chain_id`?

And then, what's up with `chain_number`?  You loop over values of it, but then don't use it anywhere in the loop.  But then you do refer to it down at line 780:

```Python
                if residue.chain.index != chain_number:
                    continue  # Only modify specified chain
```

At that point, `chain_number` will simply be the value from the last iteration of the loop.  That is, the total number of chains.

I'm confused!

jchodera:
Whoops, good catch.  I did not fully eliminate the old weird implementation.  Just a sec.

jchodera:
Sorry about that.  The code was a mess.  Cleaned things up significantly.

peastman:
It still seems to be doing unnecessary work:

```Python
        for (chain_number, chain) in enumerate(self.topology.chains()):
            if chain.id == chain_id:
                for (residue_number, residue) in enumerate(chain.residues()):
                    resSeq_to_residue[int(residue.id)] = residue
```

You now have *two* loop variables that are never used.  That can be simplified to just


```Python
        for chain in self.topology.chains():
            if chain.id == chain_id:
                for residue in chain.residues():
                    resSeq_to_residue[int(residue.id)] = residue
```

which is a lot clearer.

jchodera:
Sure, thanks.

jchodera:
Sorry again---I didn't write the initial version of this.

peastman:
Ok, that looks better.  Thanks!

PR #115, opened 2016-01-19
Specify water box dimensions with JS enhanced padding field

jaimergp:
Instead of letting the user input each box side manually, a JS enhanced 4th box can fill those three fields automatically. 

waterbox

### DIFF
--- a/pdbfixer/html/addHydrogens.html
+++ b/pdbfixer/html/addHydrogens.html
@@ -23,6 +23,13 @@ function validateForm() {
     }
     return true
 }
+function applyPaddingToWaterBox(padding) {
+    padding = parseFloat(padding);
+    if(!padding || padding<0) padding = 0;
+    for (var i = pbcxyz.length; i--;) {
+        pbcxyz[i].value = Math.ceil((padding+molxyz[i])*1000) / 1000;
+    }
+}   
 </script>
 <form id="mainform" method="post" action="/">
 <h1>Delete Heterogens</h1>
@@ -43,7 +50,14 @@ Add a water box surrounding the model?
 <input type="checkbox" id="addWaterCheckbox" name="addwater" onchange="enableControls(document.getElementById('waterInputs'), document.getElementById('addWaterCheckbox').checked)"> Add water
 <div id="waterInputs" style="margin-left:50px">
     <table style="text-align:right">
-        <tr><td>Box dimensions (in nm):</td><td><input type="text" id="boxxfield" name="boxx" size="8"></td><td><input type="text" id="boxyfield" name="boxy" size="8"></td><td><input type="text" id="boxzfield" name="boxz" size="8"></td></tr>
+        <tr>
+        <td>Box dimensions (in nm):</td>
+        <td><input type="text" id="boxxfield" name="boxx" size="8"></td>
+        <td><input type="text" id="boxyfield" name="boxy" size="8"></td>
+        <td><input type="text" id="boxzfield" name="boxz" size="8"></td>
+        <td>Padding:</td>
+        <td><input type="text" id="xyzpadding" name="xyzpad" size="8" oninput="applyPaddingToWaterBox(this.value)"></td>
+        </tr>
         %s
     </table>
     <p>
@@ -59,8 +73,10 @@ Add a water box surrounding the model?
 <input type="button" value="Continue" onclick="if (validateForm()) submitWithSpinner()"/>
 </form>
 <script>
-setCurrentStep(6)
-enableControls(document.getElementById('waterInputs'), false)
+setCurrentStep(6);
+enableControls(document.getElementById('waterInputs'), false);
+var molxyz = [].slice.call(document.getElementById('waterInputs').getElementsByTagName('tr')[1].getElementsByTagName('td')).splice(1,3).map(function(td){ return parseFloat(td.textContent)});
+var pbcxyz = [document.getElementById('boxxfield'), document.getElementById('boxyfield'), document.getElementById('boxzfield')];
 </script>
 </body>
 <html>
--- a/pdbfixer/html/addHydrogens.html
+++ b/pdbfixer/html/addHydrogens.html
@@ -26,10 +26,18 @@ function validateForm() {
 function applyPaddingToWaterBox(padding) {
     padding = parseFloat(padding);
     if(!padding || padding<0) padding = 0;
-    for (var i = pbcxyz.length; i--;) {
-        pbcxyz[i].value = Math.ceil((padding+molxyz[i])*1000) / 1000;
+    if (document.getElementById('cubicPaddingCheckBox').checked){
+        var maxAxis = Math.max.apply(Math, molxyz);
+        for (var i = pbcxyz.length; i--;) {
+            pbcxyz[i].value = Math.ceil((padding+maxAxis)*1000) / 1000;
+        }
     }
-}   
+    else {
+        for (var i = pbcxyz.length; i--;) {
+            pbcxyz[i].value = Math.ceil((padding+molxyz[i])*1000) / 1000;
+        }
+    }
+}    
 </script>
 <form id="mainform" method="post" action="/">
 <h1>Delete Heterogens</h1>
@@ -51,12 +59,15 @@ Add a water box surrounding the model?
 <div id="waterInputs" style="margin-left:50px">
     <table style="text-align:right">
         <tr>
-        <td>Box dimensions (in nm):</td>
+        <td>Type box dimensions (in nm):</td>
         <td><input type="text" id="boxxfield" name="boxx" size="8"></td>
         <td><input type="text" id="boxyfield" name="boxy" size="8"></td>
         <td><input type="text" id="boxzfield" name="boxz" size="8"></td>
-        <td>Padding:</td>
+        </tr>
+        <tr>
+        <td>... or set them from padding:</td>
         <td><input type="text" id="xyzpadding" name="xyzpad" size="8" oninput="applyPaddingToWaterBox(this.value)"></td>
+        <td colspan="2"><input type="checkbox" id="cubicPaddingCheckBox" name="cubicPaddingCheckBox" checked>Force cubic geometry <span title="All sides are equal to the biggest dimension plus padding" style="border-bottom: 1px dotted black; cursor: help">?</span></td>
         </tr>
         %s
     </table>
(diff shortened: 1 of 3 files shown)


peastman:
This behavior is different from what `Modeller.addSolvent()` does if you specify the `padding` argument.  That takes the largest dimension of the box, adds the padding to that, then sets all three dimensions to the result so you get a cubic box.  The idea is that if you're specifying a padding, it's usually because you're simulating something in solvent (rather than a crystal), so you want to make sure however it drifts or rotates, there will always be a sufficient distance between periodic copies.  So I'd suggest changing this to work the same way.  Do other people agree?

peastman:
To avoid confusion, it might be good to also make it clear the padding box if you type anything into the X, Y, or Z box.  And possibly change the label to "Set from padding:".  We want to make it clear to people that they don't need to enter both box dimensions *and* a padding distance.

jaimergp:
Of course, if that's Modeller behaviour, it should be reflected here as well! I am no expert in MD, so yes, that seems very reasonable.

A middle ground would consist of an additional checkbox "Force cubic geometry", enabled by default, possibly with one of those question-mark floating tips, explaining why it's enabled by default.

That way, the user can choose anytime?

peastman:
Of course, "force cubic geometry" would only apply to the "padding" box.  If users entered box dimensions directly, those wouldn't need to be cubic.  We'd need some way to make that clear.

Perhaps people who actually run simulations (that is to say, not me!) could comment on which behaviors are most useful?

swails:
I rarely care to use a cubic box.  I'd much sooner use a truncated octahedron or rhombic dodecahedron if I wanted all of my unit cell sides to be the same size (which would ensure enough space to tumble).

The only time I really use orthogonal cells is when I'm doing either a crystal simulation with an orthorhombic crystal, in cases where a box with equal sides is simply too large to simulate (e.g., for a very long system), [1] or with surfaces (like bilayer simulations), where the box size is set by the surface added to the system.

[1] I've never actually done a simulation like that, but I know colleagues in groups I've worked with that have.

peastman:
Perhaps if the user sets the box based on padding, there should be a set of choices for box shape: rectangle, cube, truncated octahedron, or rhombic dodecahedron?

jchodera:
I like cubic boxes!

swails:
> I like cubic boxes!

That's like using an olympic-regulation swimming pool as a hot tub :).

jaimergp:
I implemented @peastman first suggestion:

captura de pantalla de 2016-01-19 22-02-13

With respect to the other suggestion, that's way out of this PR, right? It would need a backend implementation first, maybe in Modeller? Then, yes, a dropdown list would do the trick. Something like this:

captura de pantalla de 2016-01-19 22-14-08

peastman:
`Modeller.addSolvent()` already has a `boxVectors` parameter that you can use to specify non-rectangular boxes.  We need to add that option to `PdbFixer.addSolvent()`, but that's trivial.  It just needs to pass it on to `Modeller` the same way it does for the other options.

Does anyone have a reference for how to compute the box vectors given the desired shape, the box size, and the padding?

jaimergp:
Oh, I see, didn't know about that! That's great! 

I know Gromacs can handle three box types (cubic, rhombic dodecahedral, or truncated octahedral box) with [editconf](http://manual.gromacs.org/programs/gmx-editconf.html). Maybe the [source](https://github.com/gromacs/gromacs/blob/master/src/gromacs/gmxana/gmx_editconf.cpp) its helpful.

peastman:
Even better, the Gromacs manual lists exactly what the box vectors should be.  See table 3.1 in http://manual.gromacs.org/documentation/5.1.1/manual-5.1.1.pdf.  They give vectors for two different orientations of the rhombic dodecahedron.  The "xy-square" one is what editconf produces, so it probably makes sense for us to use that too.

jaimergp:
[Chimera's MetalGeom](http://svn.cgl.ucsf.edu/svn/chimera/branches/qtdemo/libs/MetalGeom/geomData.py) has a decent collection of polyhedrons with their normal vectors too. I mean, A LOT:

- hexagonal bipyramid
- octahedron, trans-bicapped
- icosahedron
- monocapped square antiprism
- tetrahedral
- bisdisphenoid
- trigonal prism
- trigonal prism, square face-bicapped
- anticuboctahedron
- trigonal prism, square face-tricapped
- cube
- trigonal bipyramid
- face-capped octahedron
- pentagonal bipyramid
- octahedron
- cuboctahedron
- square pyramid
- bicapped hexagonal antiprism
- square planar
- square antiprism
- trigonal prism, square face monocapped
- trigonal prism, triangular face-bicapped
- bicapped square antiprism

peastman:
I think that might be a *little* excessive for our purposes. :)  Not that it wouldn't be cool to offer a monocapped square antiprism.  I mean, I don't know what it is, but the name has a nice sound!

peastman:
@jaimergp do you want to attempt these changes?  If you don't want to that's fine.  I can do it and then send them as a PR to your repository so they then get added to this PR automatically.

jaimergp:
Hi @peastman,
I have tried to implement it myself, but Modeller complains about the vectors. I've  attached an IPython notebook with my failed attempt. I hope it's clear!

[New solvation boxes for PDBFixer.ipynb.zip](https://github.com/pandegroup/pdbfixer/files/99867/New.solvation.boxes.for.PDBFixer.ipynb.zip)

peastman:
The table lists the vectors going down the columns, not across the rows.  So for rhombic dodecahedron, the first vector (labeled "a" in the table) is (1, 0, 0), not (1, 0, 0.5).

jaimergp:
I feel veeery dumb right now. The snippet runs now, but I don't know if the outputs are right. Aren't the boxes supposed to resemble their respective polyhedron? I mean, the file sizes and the occupied volume are in line with the expected values, but I am not seeing a rhombic dodecahedron or a truncated octahedron anywhere... I'll leave an attachment with the boxes and the snippet, and will check tomorrow again! Thanks, by the way, @peastman. I am learning a lot.

[failure_num2.zip](https://github.com/pandegroup/pdbfixer/files/99989/failure_num2.zip)

peastman:
No, they don't actually look like them.  That's the cool thing: it turns out that each of those box shapes is mathematically identical to a triclinic box of the right size and shape.  So as long as you support triclinic boxes, you get all those different box shapes for free.  This is the paper that introduced the idea: http://www.rug.nl/research/portal/files/3218542/1997JComputChemBekker.pdf.

swails:
Exactly.  The anisotropies in the box vectors mean that the "closest" images in the corners of the cells are not actually the closest images to the "center" of the unit cell.

So if you cut those corners off, and put them on the opposite side of the box, you get a more spherical representation.

These ideas are formalized in the paper @peastman linked, but hopefully it provides a more visual description.

jaimergp:
A-ha.... Figure 8 in the paper made it clear! Out of curiosity, how would a user "reconstruct"  the intended geometry for depiction purposes? Are programs like VMD capable of such task?

/offtopic

I'll proceed to implement the snippet in the PR during the weekend. Thanks again!

swails:
> how would a user "reconstruct" the intended geometry for depiction purposes?

I'm not sure what you mean by "intended geometry".  I *think* you mean that it takes the most spherical geometry possible.  In any case, to create the "triclinic" shape, just apply the standard [fractional coordinate truncation](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/distance.py#L227-L237) (particularly line 234 which does the truncation), then map the fractional coordinates back to Cartesian space.

To get the "familiar" shape (like the truncated octahedron or rhombic dodecahedron), you take the "triclinic" shape and for each atom find the image that is *closest* to the center of the unit cell.  Because orthorhombic cells are isotropic, every particle in a unit cell is already the closest image possible to the unit cell center.  For non-orthorhombic cells, what [mdtraj currently does](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/distance.py#L238-L248) is search the 26 "adjacent" unit cells and see if for any atom there is an image that is *closer* to the unit cell center than the "standard" image in the triclinic cell.

However, as @rmcgibbo pointed out in mdtraj/mdtraj#1019 and cpptraj/cpptraj#223, this approach does not work for highly-skewed unit cells.  You can imagine taking a unit cell with *incredibly* acute angles, and make the "closest" image an arbitrary number of unit cells away from the triclinic image (as the angle between the UC vectors approaches 0).  But these are highly unusual (which is to say, I've never seen them before) in standard MD applications.  

And "familiar" imaging requires 27 distance evaluations instead of just 1 if you only search unit cells that are "one" away from the central unit cell.  Going outward one more unit cell (which will correctly image for slightly more skewed boxes) balloons that number to 125.  Which is why cpptraj and mdtraj have those reported bugs in the first place, and [GROMACS simply warns that the box is skewed and knowingly turns a potentially-incorrect minimum image distance](https://github.com/Amber-MD/cpptraj/issues/223#issuecomment-170085799).

Thus ends probably more than you wanted to know about PBC, imaging, and distances :).

jaimergp:
Ok, hopefully that should do it! The attachment contains test PDB files for each geometry.

[1rfo_solvation_boxes.zip](https://github.com/pandegroup/pdbfixer/files/106403/1rfo_solvation_boxes.zip)

(By the way, @swails, I thought I replied to you but it's obvious I didn't... Thanks for the thorough explanations!)

peastman:
I realized this never got merged.  Sorry about that!  I just tested it, and it mostly works fine for me.  I did find one thing that isn't working correctly: there's a ? after the choice for box shape that's clearly intended to be a help link, but nothing happens when I click on it.  I tried in both Firefox and Safari, and it's the same in both.

jaimergp:
Hey! No problem at all :)

The "?" is not clickable. The user is supposed to hover the cursor and after a couple of seconds the standard tooltip/hint will appear (like title attributes in images). Would you prefer a clickable link?

peastman:
Well, I assumed I was supposed to click it.  And when nothing happened, I concluded it was broken.  So we have at least one data point of a user who was confused by the current behavior. :)

jaimergp:
But with `n=2`... we need a larger sample! :D I'll have a look and see if something sexy is achievable with CSS only.

jaimergp:
I forgot to add, this commit doesn't make the link clickable, but the tooltip appears instantaneously, so I guess there's no possible mistake now.

peastman:
That works.  Ok, I'll go ahead and merge this.  I think it's a really great new feature.  Thanks!

PR #120, opened 2016-03-22
Fixed some failing test cases

peastman:


peastman:
All tests passed, so I'm going to merge.  However, travis is still showing the build as failing.  I think it's because of this error:

```
$ pyflakes pdbfixer/*.py
/home/travis/build.sh: line 45: pyflakes: command not found
```

Do we need to modify the travis script to install that?

PR #124, opened 2016-03-23
Removed some unused code

peastman:


peastman:
Something really strange is going on.  For Python 3.4 it's still reporting the error in

    File "/home/travis/miniconda/envs/_test/lib/python3.4/site-packages/pdbfixer/createSoftForcefield.py", line 54, in <module>

But there's no such file anymore!  `createSoftForcefield.py` is now in the `devtools` directory.  Has travis somehow cached it somewhere?

jchodera:
It looks like it's [installing a conda package called `pdbfixer-dev-dev`](https://travis-ci.org/pandegroup/pdbfixer/jobs/118047104#L547-L553
):
```
TEST START: pdbfixer-dev-dev-py34_0
Fetching package metadata: ........
Solving package specifications: .........
The following packages will be downloaded:
    package                    |            build
    ---------------------------|-----------------
    nose-1.3.7                 |           py34_0         195 KB
    pyflakes-1.1.0             |           py34_0          69 KB
    pdbfixer-dev-dev           |           py34_0         182 KB
    ------------------------------------------------------------
                                           Total:         445 KB
```
This package [was built six months ago](https://anaconda.org/omnia/pdbfixer-dev/files).

peastman:
And just before that:

```
$ conda install --yes --quiet nose nose-timer
Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata: ......
Solving package specifications: .........
Warning: 2 possible package resolutions (only showing differing packages):
  - pdbfixer-dev-dev-nppy34_0.tar.bz2
  - pdbfixer-dev-dev-py34_0.tar.bz2
```

So it seems to think that pdbfixer-dev is a dependency for either nose or nose-timer???

jchodera:
Weird.  `pdbfixer-dev` isn't a dependency for `nose-timer` (and probably not for `nose`).

This might be a weird situation where we have to delete old `pdbfixer-dev` packages.  I've just removed the [old ones from here](https://anaconda.org/omnia/pdbfixer-dev/files).  Any others hiding about?

peastman:
I restarted the builds.  Let's see what happens.

peastman:
That seems to have fixed the error.  But it still downloaded pdbfixer-dev, it just got a newer version of it.

I'll go ahead and merge this.  Then I'll send a separate PR removing pyflakes, at which point travis should start showing the builds as passing.  But we still need to figure out what's going on with pdbfixer-dev.

mpharrigan:
Why are you running the tests as seperate lines in .travis.yml? You can use conda build to automatically download test requirements and run tests in a special test environment. Check out how we do msmbuilder and mdtraj. The .travis.yml only does "conda build", and the recipe contains the call to nosetests, et al.

I suspect this would fix the issue of it pulling in the internet version of pdbfixer-dev instead of the just-built version of pdbfixer-dev

jchodera:
> Why are you running the tests as seperate lines in .travis.yml? 

This was a decision made some time ago because we wanted to have more control over testing than was possible in the very limited conda recipe test environment (e.g. we'd have to wrap `nose-timer` as a conda recipe!) and we wanted to avoid testing twice.

> I suspect this would fix the issue of it pulling in the internet version of pdbfixer-dev instead of the just-built version of pdbfixer-dev

Not certain how. We use the `_test` environment created by `conda build` to do the testing.

mpharrigan:
> (e.g. we'd have to wrap nose-timer as a conda recipe!)

`nose-timer` seems to already be a conda package, from travis.yml:
```
conda install --yes --quiet nose nose-timer
```

>  we wanted to have more control over testing than was possible in the very limited conda recipe test environment

You can install any conda package and run arbitrary shell commands. Seems like it would work well here

mpharrigan:
> Not certain how. We use the _test environment created by conda build to do the testing.

Yeah, it's confusing; but relying on an environment named "_test" seems improper, given python's naming scheme of underscore-prefixed variables to be treated as private/implementation details

jchodera:
> Yeah, it's confusing; but relying on an environment named "_test" seems improper, given python's naming scheme of underscore-prefixed variables to be treated as private/implementation details

There are other ways to do this by forcing `conda` to install locally, but these sometimes led to the wrong version being installed, no version being installed, or the installation taking a lot longer because many dependencies had to be re-installed.  This approach seems to be a useful compromise for the time being, since the `_test` environment is preconfigured with everything we want.

The biggest open question is whether all the testing should be done in the conda recipe `test:` block instead of the `.travis.yml` script block. We had done things in the script block because it gave us more flexibility to install some things (like openeye tools) outside of the weird conda controlled build/test environment.

PR #130, opened 2016-04-13
Updated version number to 1.3

peastman:


jchodera:
I think you said that we need `openmm >=7.0` features in here, right? So you need to update the dependencies too:
https://github.com/pandegroup/pdbfixer/pull/130/files#diff-2eeaed663bd0d25b7e608891384b7298R77

peastman:
Good point.  Fixed!

jchodera:
You want to cut a new GitHub release, and then I'll update the conda recipe?

peastman:
Release created: https://github.com/pandegroup/pdbfixer/releases/tag/v1.3

jchodera:
recipe in test: https://github.com/omnia-md/conda-recipes/pull/493

PR #143, opened 2016-12-05
Fixed code that was broken by changes to OpenMM

peastman:


peastman:
Travis is failing with the error message, "Miniconda MD5 mismatch".  Any idea what's causing that?

mpharrigan:
https://github.com/msmbuilder/msmbuilder/pull/933

change http to https when downloading miniconda

peastman:
Thanks!

peastman:
Now it's failing with this error:

```
Can't build /home/travis/build/pandegroup/pdbfixer/devtools/conda-recipe due to unsatisfiable dependencies:

Package missing in current linux-64 channels: 

  - openmm-dev
```

I guess that's because we got rid of the `openmm-dev` package, and it should instead be using the `dev` label of the `openmm` package.  How do I specify that?  I've been looking through the documentation at http://conda.pydata.org/docs/building/meta-yaml.html#requirements-section, but I can't find any mention of how to specify labels.

jchodera:
We've deprecated the `openmm-dev` package, and in this specific case, it's best to make sure nobody is continuing to still use it.

Here's what I would do:
* Change your `.travis.yaml` to add after [this line](https://github.com/pandegroup/pdbfixer/blob/master/.travis.yml#L15):
```YAML
  - conda config --add channels $ORGNAME/labels/dev
```
* Change the [`openmm-dev` dependencies in `meta.yaml`](https://github.com/pandegroup/pdbfixer/blob/master/devtools/conda-recipe/meta.yaml#L13-L18) to just `openmm`
* Enable nightly builds with [`nightli.es`](https://nightli.es/) so we don't miss changes that break `pdbfixer`

peastman:
It doesn't work.  The added line fails with this error message:

```
conda.exceptions.CondaHTTPError: HTTP 404 None

for url <None>

The remote server could not find the channel you requested.
```

jchodera:
Weird. Maybe $ORGNAME is incorrect?
```bash
conda config --add channels omnia/labels/dev
```
works fine for me. Try it on your terminal?

jchodera:
Whoops, I see the issue.

jchodera:
It has to be 
```bash
conda config --add channels https://conda.anaconda.org/omnia/label/dev
```
Sorry about that!

peastman:
Thanks!  Another inch forward.  Now it's failing with this error:

```
$ source activate _test

CondaEnvironmentNotFoundError: Could not find environment: _test .

You can list all discoverable environments with `conda info --envs`.
```

Sure enough, the line

```
  - source activate _test
```

appears in .travis.yml, but I have no idea what it's supposed to be doing.  And all of this used to work.  I'm really confused about what's going on.

mpharrigan:
Ha! I told you not to do it that way because it might change :)

peastman:
Yay!  That finally worked.

jchodera:
It was much more elegant than the current approach...

jchodera:
...even if short-lived.

mpharrigan:
At the risk of fixing something that's working, why can't you just put the test command in the conda recipe and have the tests run automatically during conda build? This is how we do mdtraj and msmbuilder [although to build the docs, we have to create an env, use install --local, etc]

jchodera:
Many of our projects have to also test optional dependencies on other tools, like the OpenEye toolkit, which is only installable through `pip` and needs secure license file dependencies.

PR #174, opened 2018-09-14
Update usage as "prog filename [options]"

T-B-F:
filename must be the first argument of pdbfixer when used with options (line 1218)


peastman:
Thanks!  Good catch.

Issue #176, opened 2018-10-03
Non-reproducible behaviour of `addMissingAtoms`


JoaoRodrigues:
Running `addMissingAtoms()` gives different results because the `LocalEnergyMinimizer` does not allow setting a random seed from `pdbfixer`. The end result is that re-built sidechains are pointing in different directions at every different try, which is not good for reproducibilty. Setting `random.seed` and `np.random.seed` does not help. The following code snippet ran on any structure missing heavy atoms will reproduce the problem.

```
for i in range(5):
    random.seed(917)

    import pdbfixer as pf
    from simtk.openmm.app import PDBFile

    mol = pf.PDBFixer('../example/4xt1_AB.pdb')
    mol.findMissingResidues()
    mol.findMissingAtoms()
    mol.addMissingAtoms()

    with open('output_{}.pdb'.format(i), 'w') as handle:
        PDBFile.writeFile(mol.topology, mol.positions, handle)
```

My suggestion would be add an optional seed parameter to `addMissingAtoms` that then sets the integrator seed. I'll follow with a pull request.

PR #177, opened 2018-10-03
Added a seed parameter to addMissingAtoms to allow reproducible results

JoaoRodrigues:
Addresses issue #176. Setting the random seed when calling `addMissingAtoms` yields reproducible atom positions, while no seed results in the current behaviour (no reproducibility).

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -855,8 +855,14 @@ class PDBFixer(object):
         self.missingAtoms = missingAtoms
         self.missingTerminals = missingTerminals
 
-    def addMissingAtoms(self):
+    def addMissingAtoms(self, seed=None):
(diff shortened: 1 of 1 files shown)


peastman:
Looks good.  Be aware though that even if you set the seed, that doesn't guarantee you'll get identical results every time.  It depends on the platform you're using.  For example, on the CPU platform the CustomNonbondedForce sums interactions in a non-deterministic order.  That causes the initial forces to be slightly different, which can cause the minimization to end up somewhere different.

JoaoRodrigues:
Thanks Peter. Indeed, but at least on the same platform running the same script gives the same result.

I'd consider actually adding a similar setting for `addHydrogens` in openmm because of a similar issue. Maybe there a global `setSeed` method would be better, for simplicity.

Issue #175, opened 2018-10-03
pdbfixer is mislabeling built residues with negative res IDs


nitroamos:
In this [line](https://github.com/pandegroup/pdbfixer/blob/699ba1033a46c2a862f4cce0f271c37667677f07/pdbfixer/pdbfixer.py#L513)

            newResidue = chain.topology.addResidue(residueName, chain, "%d" % ((firstIndex+i)%10000))

PDBFixer is wrapping negative residue numbers around 10000, meaning that a residue whose number is supposed to be -4 is ending up as 9996.

One fix would look like this:

            newResId = firstIndex+i
            if len(str(newResId)) >= 5:
		   newResId = (firstIndex+i)%10000
            newResidue = chain.topology.addResidue(residueName, chain, "%d" % (newResId))

which is closer to what happens in [OpenMM](https://github.com/pandegroup/openmm/blob/6fb038b99516a26c5349dd08343f6137b96f8663/wrappers/python/simtk/openmm/app/pdbfile.py#L353)

Or even simpler would be to not do the modulo in PDBFixer since OpenMM does it.

peastman:
Where did you find a PDB file with negative residue numbers?  Residue numbers are supposed to be the index within the SEQRES section, which by definition can never be negative.

nitroamos:
In my test case, it's coming from a `REMARK 465` section which is integrated with PDBFixer as outlined [here](https://github.com/pandegroup/pdbfixer/issues/170#issuecomment-410722072). For example, here's a random one Google found for me, take a look [here](https://files.rcsb.org/header/1P0R.pdb)
```
REMARK 465                                                                      
REMARK 465 MISSING RESIDUES                                                     
REMARK 465 THE FOLLOWING RESIDUES WERE NOT LOCATED IN THE                       
REMARK 465 EXPERIMENT. (RES=RESIDUE NAME; C=CHAIN IDENTIFIER;                   
REMARK 465 SSSEQ=SEQUENCE NUMBER; I=INSERTION CODE.)                            
REMARK 465     RES C SSSEQI                                                     
REMARK 465     MET A   -19                                                      
REMARK 465     GLY A   -18                                                      
REMARK 465     SER A   -17                                                      
REMARK 465     SER A   -16
...
```

 I think the scientific origin of this is when people want to number their residues based on a pre-existing sequence alignment.

peastman:
Good point.  So PDBFixer really doesn't need to do this.

Issue #170, opened 2018-06-04
More rigorous implementation of findMissingResidues needed (e.g., doesn't work for antibodies)


nitroamos:
While writing up my response to https://github.com/pandegroup/openmm/issues/2087 I went poking around in `findMissingResidues` to figure out how it's been implemented, and I found a problem.

First, to recap that discussion:
1) You can't rely on residue numbers alone to determine where gaps in the structures are. Residues numbers are often chosen on the basis of an alignment with a family of proteins and as such can jump up or down. If you want to take an approach like this, you should use `r.index` instead of `r.id`. [Example.](https://biology.stackexchange.com/questions/39122/do-residue-sequence-numbers-in-pdb-files-correspond-to-the-positions-in-the-back)

2) You can't rely on `SEQRES` to find gaps in the structure. For example, if `SEQRES` says the sequence is `AAA` and the coordinate section says `AA`, there are 3 places you could put the missing A. 

3) I'm not a PDB historian so I don't know why it's in a REMARK section, but [REMARK 465](https://www.wwpdb.org/documentation/file-format-content/format33/remarks2.html#REMARK%20465) is supposed to be where you look to find where the missing residues are and given the template they provide, it is intended to be machine readable.

I went looking in the code for `findMissingResidues` to understand it. As far as I can tell, it makes these assumptions:
1) it does not use `REMARK 465`. This means it could be fooled by `SEQRES` resulting in misthreaded structures.
2) By not using `REMARK 465`, it doesn't know if the missing residues are already named (i.e., id and insertionCode)
3) it assumes residue numbers are sequential
4) It doesn't provide any warnings if something looks suspicious.

Take a look [at this line](https://github.com/pandegroup/pdbfixer/blob/6c1caa963b6423ed8412ee539fa2ef77a5fe69d5/pdbfixer/pdbfixer.py#L597): 

```
        for chain in chains:
            minResidue = min(int(r.id) for r in chain.residues())
            maxResidue = max(int(r.id) for r in chain.residues())
            residues = [None]*(maxResidue-minResidue+1)
            for r in chain.residues():
                residues[int(r.id)-minResidue] = r.name
            chainWithGaps[chain] = residues
```

[...]

PR #178, opened 2018-10-04
Some bug fixes related to building missing residues

nitroamos:
Some of my changes are related to [this](https://github.com/pandegroup/pdbfixer/issues/170) where I want to be able to specify the missing residues using `REMARK 465` sections. For now, I create `chainWithGapsOverride` in my own software (an old version is in issue #170) because it's quite complex and because I'm not sure if/how PDBFixer/OpenMM could/should extract and save the `REMARK 465` section from the original pdb.

I did make one change to `chainWithGaps` corresponding to the situation where there are waters following the sequence -- their presence was preventing the "alignment" step in `findMissingResidues` from working. I obviously don't have a list of all possible ligands and I don't want to examine CONECT records, but the approach of "known" residues seems ok to me.

This addresses an unreported issue of insertionCodes for not-missing residues are not preserved.

One problem I've not taken on is that `REMARK 465`'s specified resnum and insertion code are not necessarily preserved by `_addMissingResiduesToChain`. Seems minor and I don't have any tests cases for that right now.

This also addresses #175

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -510,7 +510,7 @@ class PDBFixer(object):
 
             # Create the new residue.
 
-            newResidue = chain.topology.addResidue(residueName, chain, "%d" % ((firstIndex+i)%10000))
+            newResidue = chain.topology.addResidue(residueName, chain, "%d" % (firstIndex+i))
             fraction = (i+1.0)/(numResidues+1.0)
             translate = startPosition + (endPosition-startPosition)*fraction + loopHeight*math.sin(fraction*math.pi)*loopDirection
             templateAtoms = list(template.topology.atoms())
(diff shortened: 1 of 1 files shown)


JoaoRodrigues:
Just to add that I am running into a similar issue and a patch like this _could_ be potentially very useful. In my situtation, PDBFixer cannot add caps to the middle of a structure (in case it has gaps, this is sometimes useful before dynamics).

I cannot say this is a perfect solution, would need more time to check it thoroughly, but just adding my 2 cents that this is an important issue.

peastman:
Sorry I've been so bad about replying to this.  I kept meaning to reply but not doing it, and then I was traveling, and then I was on vacation.  Sorry!

Anyway, I agree with you.  This is an important issue, but I'm not convinced this is exactly the right solution.  Could you open an issue describing exactly what you see as the important issues and requirements?  Let's discuss what is the best way of solving it.

Issue #145, opened 2017-01-10
N-terminal residue added with wrong chirality and/or a cis-peptide bond


rafwiewiora:
We've been on a chirality and cis-peptide mistake hunting recently, after we discovered that `Ensembler` was causing these in some simulations. I also just looked through some of my manually set up simulations and discovered that in some cases `pdbfixer` had modeled in the N-terminal residue with the wrong chirality and with a cis-peptide bond to the next residue. 

Here's an example (turns out I had `pdbfixer 1.2` installed when I'd run this):
(`pdbfixer                  1.2                      py27_1    omnia`)

Starting structure: 2BQZ

Code:
``` python
import pdbfixer
from simtk.openmm.app import PDBFile

fixer_SET8 = pdbfixer.PDBFixer('pdb2bqz.ent')
fixer_SET8.removeChains([1,2,3,4,5,6,7,8,9])
fixer_SET8.findMissingResidues()
fixer_SET8.missingResidues = {(0,0): ['SER']}
fixer_SET8.findMissingAtoms()
fixer_SET8.addMissingAtoms()
PDBFile.writeFile(fixer_SET8.topology, fixer_SET8.positions, open('SET8_noH.pdb', 'w'))
```

Output: (SER0 as spheres, ARG1 as sticks, note both the wrong chirality of SER0 and cis-peptide bond SER0 - ARG1)

![2bqz_pdbfixer_wrong_chir_cispept](https://cloud.githubusercontent.com/assets/12088757/21819670/a8f49ae6-d73a-11e6-9ee0-b5ac389f5555.png)

Reproducing on the newest `pdbfixer`: 
(`pdbfixer                  1.3.1                    py35_1    omnia`)

I re-run the above code 10 times. We have:

* all good: 2/10
* only chirality wrong: 3
* only cis-peptide: 3
* both chirality wrong and cis-peptide: 2

We check for these problems using the VMD plugins: http://www.ks.uiuc.edu/Research/vmd/plugins/chirality/ http://www.ks.uiuc.edu/Research/vmd/plugins/cispeptide/

@maxentile has written code to do this in MDTraj, I just need to write a few tests and we will contribute. 

I wonder if `modeller.addHydrogens()` would benefit from at least adding chirality checks for the residues we protonate or even a chirality repair code? Currently the wrong chirality just goes through no problem. 

@jchodera will be interested

peastman:
[...]

PR #198, opened 2020-02-29
[WIP] Improved method of selecting positions for added residues

peastman:
This is an attempt at fixing some of the problems in #145.  After selecting initial positions for added residues, but before running the main energy minimization, it optimizes a coarse grained model that represents each residue as a single bead.  This should reduce the chance of a new residue getting put right on top of an existing one.


jchodera:
@peastman: Is this still in the works, or is it ready to merge?

peastman:
No, the changes so far didn't solve the problem.  I'm still trying to figure it out.

PR #211, opened 2020-10-12
Fix Travis builds

peastman:
See https://github.com/openmm/pdbfixer/pull/208#issuecomment-705333008.


peastman:
We're going to have to switch to a different CI system, since travis is dropping free support for open source projects.  It isn't worth spending any more effort on this PR.

PR #221, opened 2021-07-05
Fix reading PDBx from filename

### LINKED ISSUE #202: Problem with mmcif format
i'have an error: OSError: [Errno 36] File name too long when run fixer for code: 5xy3 ribosome. 
help me.

dww100:
Lines from the file where being passed to the reader rather than the file object - simple change from 
`self._initializeFromPDBx(file.read())` to `self._initializeFromPDBx(file)` fixes #202.


peastman:
Looks good, thanks!

Issue #224, opened 2021-08-28
Convert to GitHub actions to re-enable continuous integration


jchodera:
Travis CI is broken and not coming back to life.

We should upgrade this repo to use the [MolSSI Cookiecutter](https://github.com/MolSSI/cookiecutter-cms) so that we can run continuous integration again.

swails:
Done by #225

PR #225, opened 2021-08-31
Revamp PDBFixer testing

swails:
* Use GitHub Actions to run a suite of tests
* Switch from nose to pytest (nose has not been supported in years)
* Download a PDB file and track it instead of fetching it during CI
  (RCSB has blocked testing frameworks in the past).


swails:
I hadn't seen #222 

This actually adds some testing, or it should at least.

jchodera:
Also fixes #224.

We should still likely convert to using the [MolSSI cookiecutter](https://github.com/MolSSI/cookiecutter-cms) at some point down the road!

swails:
Tests now pass: https://github.com/swails/pdbfixer/runs/3468227261?check_suite_focus=true

I'll clean up the branch trigger names and this should be good to merge.  Then I'll proceed to the cleanups I need.

swails:
@peastman this is ready to merge, and results in passed tests on Python 3.7 - Python 3.9

peastman:
That looks excellent.  Thank you!

PR #228, opened 2021-09-09
Fixed error leading to duplicate bonds

peastman:
This fixes an error that would sometime cause it to produce a PDB file with duplicate CONECT records for nonstandard residues.  However, it only happened if no heavy atoms were being added.  If it needed to add even a single heavy atom anywhere it the system, it followed a different code path and the error didn't happen.


peastman:
The test failures on Python 3.7 are all from "Connection timed out".

PR #246, opened 2022-04-06
fix CI badge; remove some old simtk

dwhswenson:
The README badge didn't get updated after the switch to GitHub actions in #225. Fixes that. Also, I caught a couple places that still had `simtk` imports.


peastman:
Thanks!

PR #262, opened 2022-12-09
Add forcefield argument to addMissingHydrogens.

jonathanking:
The modeller class allows the user to specify a forcefield when adding hydrogens. This pull request simply makes this argument available to the user when using pdbfixer.

There is also a small fix of trailing whitespace.

I used this in my research when adding hydrogens to protein structures before measuring their energy with the same forcefield. Specifying the forcefield in the pdbfixer meant I would have more reasonable hydrogen placements for downstream steps.


peastman:
Looks good.  Thanks!

Issue #264, opened 2022-12-12
For those care about the running time


Minys233:
Hi researchers and developers there. Recently, I'm using PDBFixer to fix protein structures in a large scale, but I find that for large proteins, the PDBFixer runs **extremely slow**. 

After spending hours on the source code, I discovered that this is caused by the low-performance implementation of the `PDBFixer._findNearestDistance()`. For details, I used package `heartrate` to generate a code bottleneck plot, which is shown here.

<img width="1152" alt="image" src="https://user-images.githubusercontent.com/23317405/207117607-d4d353bf-40d2-4a40-83a8-8c467f9b7cf3.png">

<img width="1098" alt="image" src="https://user-images.githubusercontent.com/23317405/207118777-bc1b355e-fd7c-453a-905a-b231a9e5f558.png">

Compared to other function calls, `PDBFixer._findNearestDistance()` takes the majority of the CPU running time. the numbers between lineno. are hit times for running code. I use a case, 5mcp, a large protein which is directly download from RCSB PDB, to validate the function and the output of this piece:

```python
from pdbfixer import PDBFixer
from openmm import app
def fix_protein(pfile_in, pfile_out):
    try:
        fixer = PDBFixer(filename=str(pfile_in))
        fixer.findMissingResidues()
        fixer.findNonstandardResidues()
        fixer.replaceNonstandardResidues()
        fixer.removeHeterogens() # done by pymol
        fixer.findMissingAtoms()
        fixer.addMissingAtoms()
        # fixer.addMissingHydrogens(7.0) # this will make about 2 times slower
        with open(pfile_out, 'w') as fout:
            app.PDBFile.writeFile(fixer.topology, fixer.positions, fout)
        return None
    except Exception as e:
        return e

fix_protein('/home/minys/5mcp.pdb', './test.pdb')
```

But the problem here is simple, just find the minimum distance between `atom` and other atoms which are not in the same residue. This can be efficiently done by `KDTree` implemented in `scipy`. For a simple patch, the code goes like this. 

```python
from pdbfixer import PDBFixer
from openmm import app
import openmm.unit as unit
from scipy.spatial import KDTree
import numpy as np
import sys

def _findNearestDistance(self, context, topology, newAtoms):
    positions = context.getState(getPositions=True).getPositions(asNumpy=True).value_in_unit(unit.nanometer)
    atomResidue = [atom.residue for atom in topology.atoms()]
    kdtree = KDTree(positions[:len(atomResidue)],)
[...]

PR #267, opened 2023-02-10
Speed up _findNearestDistance

edwag:
This pull request relates to #264 

I've put together a faster implementation of the `_findNearestDistance` method. I haven't tested or benchmarked this thoroughly, but I've measured a 10-100x speedup depending on the input.

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -1197,16 +1197,24 @@ class PDBFixer(object):
 
     def _findNearestDistance(self, context, topology, newAtoms):
         """Given a set of newly added atoms, find the closest distance between one of those atoms and another atom."""
-
         positions = context.getState(getPositions=True).getPositions(asNumpy=True).value_in_unit(unit.nanometer)
-        atomResidue = [atom.residue for atom in topology.atoms()]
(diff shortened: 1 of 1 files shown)


peastman:
This implementation speeds it up by a constant factor, but it still scales as O(M\*N) where M is the number of new atoms and N is the number of total atoms.  The implementation suggested in #264 used a kd-tree to improve the scaling to O(M\*log(N)).  There's also a CellList class in Modeller we could use to improve it to O(M).  For large models, that should be a lot faster.

This implementation also uses a lot of memory.  If M=1000 and N=100,000, your `differences` and `squaredDifferences` arrays will each take 1.2 GB.  That could be limiting for large models.

Let me see if I can come up with something using CellList.

peastman:
Can you take a look at #268 and see if it looks ok to you?  How is the performance on your model?

edwag:
Thanks for such a fast and thoughtful response to this PR, and thanks also for your `_CellList` implementation! #268 looks good to me and I've left a couple of comments on that PR. In my application, performance is massively improved over the old implementation, though I haven't properly benchmarked it so I can't give a quantitative answer.

edwag:
Closing because #264 is better addressed by #268

PR #268, opened 2023-02-13
Optimizations to _findNearestDistance()

### LINKED ISSUE #264: For those care about the running time

peastman:
Fixes #264.  This uses the `_CellList` class to find the neighbors of an atom in constant time.  When working with large models, this produces a major speedup.

I also made one other change.  It stops looping once the nearest pair of atoms is at least 0.13 nm apart.  It ignored pairs of atoms within the same residue, but it didn't ignore pairs in different residues that were bonded to each other (e.g. peptide bonds).  Since those may reasonably be close together, that could lead it to perform more iterations than necessary.  It now ignores all bonded pairs when finding the nearest distance.

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -953,15 +953,18 @@ class PDBFixer(object):
             context.setPositions(newPositions)
             mm.LocalEnergyMinimizer.minimize(context)
             state = context.getState(getPositions=True)
-            cutoff = 0.13
-            exclusions = dict((atom, set([a.index for a in atom.residue.atoms()])) for atom in newAtoms)
-            for a1, a2 in newTopology.bonds():
-                if a1 in exclusions:
-                    exclusions[a1].add(a2.index)
-                if a2 in exclusions:
-                    exclusions[a2].add(a1.index)
             if newTopology.getNumResidues() > 1:
-                nearest = self._findNearestDistance(context, newTopology, newAtoms, cutoff, exclusions)
+                # When looking for pairs of atoms that are too close to each other, exclude pairs that
+                # are in the same residue or are directly bonded to each other.
+
+                exclusions = dict((atom, {a.index for a in atom.residue.atoms()}) for atom in newAtoms)
+                for a1, a2 in newTopology.bonds():
+                    if a1 in exclusions:
+                        exclusions[a1].add(a2.index)
+                    if a2 in exclusions:
+                        exclusions[a2].add(a1.index)
+                cutoff = 0.13
+                nearest = self._findNearestDistance(context, newAtoms, cutoff, exclusions)
                 if nearest < cutoff:
 
                     # Some atoms are very close together.  Run some dynamics while slowly increasing the strength of the
@@ -970,7 +973,7 @@ class PDBFixer(object):
                     for i in range(10):
                         context.setParameter('C', 0.15*(i+1))
                         integrator.step(200)
-                        d = self._findNearestDistance(context, newTopology, newAtoms, cutoff, exclusions)
+                        d = self._findNearestDistance(context, newAtoms, cutoff, exclusions)
                         if d > nearest:
                             nearest = d
                             state = context.getState(getPositions=True)
@@ -1202,23 +1205,23 @@ class PDBFixer(object):
                 forcefield._templateSignatures[signature] = [template]
         return forcefield
 
-    def _findNearestDistance(self, context, topology, newAtoms, cutoff, exclusions):
+    def _findNearestDistance(self, context, newAtoms, cutoff, exclusions):
         """Given a set of newly added atoms, find the closest distance between one of those atoms and another atom."""
 
         positions = context.getState(getPositions=True).getPositions(asNumpy=True).value_in_unit(unit.nanometer)
         boxSize = np.max(positions, axis=0)-np.min(positions, axis=0)
         boxVectors = [(boxSize[0], 0, 0), (0, boxSize[1], 0), (0, 0, boxSize[2])]
         cells = app.modeller._CellList(positions, cutoff, boxVectors, False)
-        nearest2 = sys.float_info.max
+        nearest_squared = sys.float_info.max
         for atom in newAtoms:
             excluded = exclusions[atom]
             for i in cells.neighbors(positions[atom.index]):
                 if i not in excluded:
                     p = positions[atom.index]-positions[i]
-                    dist2 = np.dot(p, p)
-                    if dist2 < nearest2:
-                        nearest2 = dist2
-        return np.sqrt(nearest2)
+                    dist_squared = np.dot(p, p)
+                    if dist_squared < nearest_squared:
+                        nearest_squared = dist_squared
+        return np.sqrt(nearest_squared)
 
 
 def main():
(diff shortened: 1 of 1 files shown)


edwag:
This code looks good to me, and behaviour seems correct. It's a little awkward for me to run a controlled benchmark for my use case, but the performance is certainly much faster than the old implementation, by at least an order of magnitude.

I have a couple of thoughts on readability/style which I will leave as comments on the diffs.

pdbfixer/pdbfixer.py
```diff
@@ -1195,18 +1202,23 @@ def _createForceField(self, newTopology, water):
                 forcefield._templateSignatures[signature] = [template]
         return forcefield
 
-    def _findNearestDistance(self, context, topology, newAtoms):
+    def _findNearestDistance(self, context, topology, newAtoms, cutoff, exclusions):
```
edwag, line 1205: The `topology` argument is no longer used and can be deleted.

pdbfixer/pdbfixer.py
```diff
@@ -1195,18 +1202,23 @@ def _createForceField(self, newTopology, water):
...
         """Given a set of newly added atoms, find the closest distance between one of those atoms and another atom."""
 
         positions = context.getState(getPositions=True).getPositions(asNumpy=True).value_in_unit(unit.nanometer)
-        atomResidue = [atom.residue for atom in topology.atoms()]
-        nearest = sys.float_info.max
+        boxSize = np.max(positions, axis=0)-np.min(positions, axis=0)
+        boxVectors = [(boxSize[0], 0, 0), (0, boxSize[1], 0), (0, 0, boxSize[2])]
+        cells = app.modeller._CellList(positions, cutoff, boxVectors, False)
+        nearest2 = sys.float_info.max
```
edwag, line 1212: It wasn't obvious to me that this means "nearest squared", particularly since in `addMissingAtoms` we do have `thing2` variables where the 2 just means 2. I would consider renaming `nearest2` and `dist2` to `nearest_squared` and `dist_squared`.

pdbfixer/pdbfixer.py
```diff
@@ -953,21 +953,28 @@ def addMissingAtoms(self, seed=None):
             context.setPositions(newPositions)
             mm.LocalEnergyMinimizer.minimize(context)
             state = context.getState(getPositions=True)
+            cutoff = 0.13
+            exclusions = dict((atom, set([a.index for a in atom.residue.atoms()])) for atom in newAtoms)
```
edwag, line 957: A couple of thoughts on this:

1.  This could be written as a dict comprehension and a set comprehension, as follows:
`exclusions = {atom: {a.index for a in atom.residue.atoms()} for atom in newAtoms}`
I would personally find this a little more readable. Tastes will obviously differ!
2. It might be worth a brief comment to explain what's going on here? Perhaps something like "When checking if any pairs of atoms are closer than the cutoff distance, ignore pairs of atoms in the same residue, as well as pairs of atoms which are bonded to each other."

pdbfixer/pdbfixer.py
```diff
@@ -953,21 +953,28 @@ def addMissingAtoms(self, seed=None):
             context.setPositions(newPositions)
             mm.LocalEnergyMinimizer.minimize(context)
             state = context.getState(getPositions=True)
```
edwag, line 955: Everything from here up to the end of the `if newTopology.getNumResidues() > 1:` block is a pretty sensible self-contained chunk of behaviour. It might make sense to pull this out into a new method? So you would just have something like `state = self._pushOverlyCloseAtomsApart(newTopology, newAtoms, integrator, context)`. Probably with a better name :)

peastman:
Thanks!  I pushed changes based on your suggestions.

Issue #287, opened 2024-04-11
Error when having more than 10 chain identifiers


FAOlivieri:
I have a system with 9 chains. After solvating with pdbfixer, I end up with 11 chains (chain ids are 1 to 11). When pdbfixer gets to the following line:

```
if len(chains) == 1:
    chains[0].id = 'A'
else:
    chains[-1].id = chr(ord(chains[-2].id)+1)
```

it throws an error, because chains[-2].id is not parseable to 1 character. I fixed it by changing the code to

```
if len(chains) == 1:
    chains[0].id = 'A'
else:
    try:
        int(chains[-2].id)
    except:
        chains[-1].id = chr(ord(chains[-2].id)+1) #what happens if chain ID is more than one character? They can be two characters
    else:
        chains[-1].id = str(int(chains[-2].id)+1)
```

and now it works properly but what exactly is this part of the code trying to achieve? As far as I understand it's assigning to the last chain the chain ID that follows the previous one, but why is that necessary? In most systems that will be true by default. Also, as per the PDB specification, chain ID can be 2 characters, so the existing code will fail on 2 letters as well.

Couldn't this whole IF be deleted?

peastman:
Chain IDs are only allowed to be a single character.  Usually they're letters.  See [the PDB spec](http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM).

FAOlivieri:
Ok. Could you clarify what the line in my post is trying to achieve?

peastman:
It increments the letter.  If the last chain is A, the next one is called B.

murfalo:
Hey Peter,

I'm running into the same issue.  This seems to occur because PDBFixer uses OpenMM's modeller.addSolvent, which in turn uses topology.addChain() in a number of locations without specifying an ID:

```python3
# ...creating solvent chain...
newChain = newTopology.addChain() # modeller.py line 571
# ...
# ...creating ion chain...
ionChain = modeller.topology.addChain() # modeller.py line 373
```

(from OpenMM modeller.py [line 571](https://github.com/openmm/openmm/blob/51a112a336db6cc0aea1c32b90d9d0a1f262cbfd/wrappers/python/openmm/app/modeller.py#L571) and [line 373](https://github.com/openmm/openmm/blob/51a112a336db6cc0aea1c32b90d9d0a1f262cbfd/wrappers/python/openmm/app/modeller.py#L373))

[...]

PR #294, opened 2024-06-21
Bugfixed chain ID assignment in addSolvent (#287)

murfalo:
How does this approach work for fixing #287?  This should assign B-Z if the previous chain ID was A-Y, and otherwise leave the ID as a numeric chain ID.

One comment:  I noticed PDBFile and PDBxFile simply use `% 26` to map numeric IDs to the alphabet.  Should we warn the user somehow that disparate chains may be assigned matching IDs in these cases?

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -1094,27 +1094,16 @@ class PDBFixer(object):
 
         modeller = app.Modeller(self.topology, self.positions)
         forcefield = self._createForceField(self.topology, True)
-        original_chains = list(modeller.topology.chains())
+        old_chains = list(modeller.topology.chains())
         modeller.addSolvent(forcefield, padding=padding, boxSize=boxSize, boxVectors=boxVectors, boxShape=boxShape, positiveIon=positiveIon, negativeIon=negativeIon, ionicStrength=ionicStrength)
         new_chains = list(modeller.topology.chains())
-        added_chain_indices = range(len(original_chains), len(new_chains))
-        assert 1 <= len(added_chain_indices) <= 2, "Expected a new solvent chain and optionally a new ion chain"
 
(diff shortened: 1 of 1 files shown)


peastman:
I think we can do something much simpler.  Remember, `writeFile()` will be assigning new chain IDs anyway.  So it doesn't really matter much what we assign here.  We could do something like this.

```python
lastid = chains[-2].id
if len(lastid) == 1 and 'A' <= lastid < 'Z':
    chains[-1].id = chr(ord(lastid)+1)
else:
    chains[-1].id = 'A'
```

We need to make this change both in `addSolvent()` and `addMembrane()`.

murfalo:
Thanks for the feedback, particularly the much simpler if case and ID assignment.  I've implemented that and also taken out the assertions, .upper(), and verbose commentary to better match existing code style.

Before porting code over to `addMembrane()`, I have two quick questions regarding your proposed method:
1. My understanding is that addSolvent will typically add two chains (solvent and ion) and addMembrane will typically add three chains (lipid, solvent, and ion).  In these cases, won't `lastid = chains[-2].id` most often refer to the ID of the newly created solvent chain, which will always be numeric?
2. In either case, do you think it's worth accounting for an edge case where the user specifies 0 ionic strength?

peastman:
Good point.  The number of chains added could be either 1 or 2 for `addSolvent()`, and either 2 or 3 for `addMembrane()`.  Maybe write a `_renameNewChains()` method that can be called in both cases?  You'll pass in the number of chains that previously existed, and it will figure out how many new ones were added and pick appropriate names for them.

murfalo:
Done!  Everything is migrated into `_renameNewChains()`, and `addSolvent()`and `addMembrane()` are updated.  How do things look?

peastman:
Looks good.  Thanks!

PR #295, opened 2024-07-06
Can add hydrogens to nonstandard residues

peastman:
Previously `addHydrogens()` only worked for standard residues.  This changes it to look up any nonstandard ones in the PDB.  If it finds them and they match, it uses the online definition to add hydrogens to them as well.

This requires the changes in https://github.com/openmm/openmm/pull/4585.


peastman:
I modified the CI configuration to use an OpenMM dev build, since it's required for this change.  We should put it back once 8.2 is released.

Issue #299, opened 2024-07-17
`PDBFixer.addSolvent` often incorrectly neutralises systems with charged ligands as though they were neutral


Yoshanuikabundi:
The `addSolvent` method uses a force field to assign charges, which it then uses to add ions that neutralise the net charge of the box. Unfortunately, the force field seems to assign formal charges of zero to many charged groups in small molecules; for example the citrate ion, which has a -3 formal charge:

```python
from openff.toolkit import Molecule

top = Molecule.from_smiles("C(C(=O)[O-])C(CC(=O)[O-])(C(=O)[O-])O").to_topology()
top.molecule(0).generate_conformers(n_conformers=1)
top.to_file("citrate.pdb")

from pdbfixer import PDBFixer
import openmm

fixer = PDBFixer("citrate.pdb")

fixer.addSolvent(
    padding=2.0 * openmm.unit.nanometer,
    ionicStrength=0.1 * openmm.unit.molar,
    boxShape="dodecahedron",
)

counts = {"NA": 0, "UNK": 0, "CL": 0, "HOH": 0}
for residue in fixer.topology.residues():
    counts[residue.name] += 1
counts # 
{'NA': 3, 'UNK': 1, 'CL': 3, 'HOH': 1457}
```

The `citrate.pdb` file generated above is available here: [citrate.pdb.zip](https://github.com/user-attachments/files/16259296/citrate.pdb.zip)

With the current `main` branch (commit 92044f3), you can also see this behavior with a PDB file like [1PO0](https://files.rcsb.org/download/1PO0.pdb), which has 2 citrate ions (each with -3 formal charge).

```python
from pdbfixer import PDBFixer
import openmm

fixer = PDBFixer("1PO0.pdb")

fixer.findMissingResidues()

# This section adds caps; leave it commented to rebuild terminal loops
chains_to_cap = {chain for chain, resi in fixer.missingResidues}
for chainidx in chains_to_cap:
    chain = [*fixer.topology.chains()][chainidx]
    last_resi = len([*chain.residues()])
    fixer.missingResidues[chainidx, 0] = ['ACE']
    fixer.missingResidues[chainidx, last_resi] = ['NME']

fixer.findNonstandardResidues()
fixer.replaceNonstandardResidues()
# fixer.removeHeterogens(keepWater=True)
fixer.findMissingAtoms()
fixer.addMissingAtoms()

fixer.addMissingHydrogens(pH=7.4)

fixer.addSolvent(
[...]

PR #301, opened 2024-07-30
Neutralise systems with formal charges from the PDB

Yoshanuikabundi:
This PR addresses #299 by:

- Refactoring code that looks up the CCD into a single private method `_downloadCCDDefinition()` that parses and caches the definition.
- Adding the private method `_downloadFormalCharges()` which gets the formal charges of the atoms in a residue from the CCD
- Using `_downloadFormalCharges()` in `_createForceField(water=True)` to provide charges only when the residue is missing from the base force field and the (non-leaving) atom names in the CCD exactly match the atom names in the PDB
- Adding a test for the above behavior

The refactor avoids re-downloading a CCD entry up to 3 times (once for protonation, once for missing heavy atoms, and a third time for formal charge) and reduces duplicate code. However, it is not essential for the other functionality and can be removed or split into its own PR if desired.

This solution works around the issues discussed in #299 where CCD residues tend to hard-code a single protonation state by only using the CCD formal charge when the force field does not provide the residue. 

My reasoning is that there are two cases: first, when the force field includes a residue, it is capable of identifying the protonation state and gives the correct net charge and so it can only introduce regressions to interfere with this. In the second case, when a residue is introduced to the force field by `_createForceField(water=True)`, the atom's partial charge is currently hard-coded to 0 and so using more information to sometimes set that to better values is appropriate.

We can break down the effect of this change (modulo bugs) like so:

1. If an atom and residue is in `amber14-all.xml`, behavior is unchanged. Otherwise:
2. If an atom and residue is in the CCD and the PDB file is consistent with that definition and the atom is charged, this change corrects solvent neutralisation.
3. If a residue is in the CCD, and the PDB file re-uses identical residue and atom names for the entire residue, and that residue is neutral in the PDB but has a formal charge in the CCD, a regression is introduced. I expect this is very rare.
4. Otherwise, behavior is unchanged.

I think 2. should be much more common than 3., so this is a step in the right direction. In particular, in the very common case that PDBFixer is used to protonate a residue (with `addMissingHydrogens()`), this change guarantees that the protonation state will be consistent with the assumed charge - either both will be taken from the CCD, or the assumed charge will be inferred from the protonation state by the force field.

I think something like an `offsetCharge` argument to `addSolvent()` and `addMembrane()` which would allow a user to correct any remaining errors in neutralising charge would be great, but this requires changes to `Modeller` and so is not included in this PR.

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -467,7 +467,7 @@ class PDBFixer(object):
             return self.templates[name]
         return None
 
-    def registerTemplate(self, topology, positions, terminal=None, formalCharges=None):
+    def registerTemplate(self, topology, positions, terminal=None):
         """Register a template for a nonstandard residue.  This allows PDBFixer to add missing residues of this type,
         to add missing atoms to existing residues, and to mutate other residues to it.
         Parameters
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -1436,6 +1436,7 @@ class PDBFixer(object):
 
         """
 
+        self.downloadCharges()
         nChains = sum(1 for _ in self.topology.chains())
         modeller = app.Modeller(self.topology, self.positions)
         forcefield = self._createForceField(self.topology, True)
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -101,12 +101,13 @@ class ModifiedResidue(object):
 
 class Template:
     """Template represents a standard residue, or a nonstandard one registered with registerTemplate()."""
-    def __init__(self, topology, positions, terminal=None):
+    def __init__(self, topology, positions, terminal=None, formalCharges=None):
         self.topology = topology
         self.positions = positions
         if terminal is None:
             terminal = [False]*topology.getNumAtoms()
         self.terminal = terminal
+        self.formalCharges = formalCharges
 
 def _guessFileFormat(file, filename):
     """Guess whether a file is PDB or PDBx/mmCIF based on its filename and contents."""
@@ -279,6 +280,9 @@ class PDBFixer(object):
         if len(atoms) == 0:
             raise Exception("Structure contains no atoms.")
 
+        # Initialise the list of formal charges
+        self.formalCharges = [None] * len(atoms)
+
         # Load the templates.
 
         self.templates = {}
@@ -354,13 +358,15 @@ class PDBFixer(object):
                 for row in modData.getRowList():
                     self.modifiedResidues.append(ModifiedResidue(row[asymIdCol], int(row[resNumCol]), row[resNameCol], row[standardResCol]))
 
-    def _getTemplate(self, name):
+    def _getTemplate(self, name, downloadIfMissing=False):
         """Return the template with a name.  If none has been registered, this will return None."""
         if name in self.templates:
             return self.templates[name]
+        if downloadIfMissing and self.downloadTemplate(name):
+            return self.templates[name]
         return None
 
-    def registerTemplate(self, topology, positions, terminal=None):
+    def registerTemplate(self, topology, positions, terminal=None, formalCharges=None):
         """Register a template for a nonstandard residue.  This allows PDBFixer to add missing residues of this type,
         to add missing atoms to existing residues, and to mutate other residues to it.
         Parameters
@@ -375,6 +381,9 @@ class PDBFixer(object):
             If this is present, it should be a list of length equal to the number of atoms in the residue.
             If an element is True, that indicates the corresponding atom should only be added to terminal
             residues.
+        formalCharges: optional list of int
+            If this is present, it should be a list of length equal to the number of atoms in the residue.
+            Each element indicates the formal charge of the corresponding atom.
         """
         residues = list(topology.residues())
         if len(residues) != 1:
@@ -383,7 +392,9 @@ class PDBFixer(object):
             raise ValueError('The number of positions does not match the number of atoms in the Topology')
         if terminal is not None and len(terminal) != topology.getNumAtoms():
             raise ValueError('The number of terminal flags does not match the number of atoms in the Topology')
-        self.templates[residues[0].name] = Template(topology, positions, terminal)
+        if formalCharges is not None and len(formalCharges) != topology.getNumAtoms():
+            raise ValueError('The number of formal charges does not match the number of atoms in the Topology')
+        self.templates[residues[0].name] = Template(topology, positions, terminal, formalCharges)
 
     def downloadTemplate(self, name):
         """Attempt to download a residue definition from the PDB and register a template for it.
@@ -419,18 +430,22 @@ class PDBFixer(object):
         xCol = atomData.getAttributeIndex('pdbx_model_Cartn_x_ideal')
         yCol = atomData.getAttributeIndex('pdbx_model_Cartn_y_ideal')
         zCol = atomData.getAttributeIndex('pdbx_model_Cartn_z_ideal')
+        zCol = atomData.getAttributeIndex('pdbx_model_Cartn_z_ideal')
+        formalChargeCol = atomData.getAttributeIndex('charge')
         topology = app.Topology()
         chain = topology.addChain()
         residue = topology.addResidue(name, chain)
         positions = []
         atomByName = {}
         terminal = []
+        formalCharges = []
         for row in atomData.getRowList():
             atomName = row[atomNameCol]
             atom = topology.addAtom(atomName, app.Element.getBySymbol(row[symbolCol]), residue)
             atomByName[atomName] = atom
             terminal.append(row[leavingCol] == 'Y')
             positions.append(mm.Vec3(float(row[xCol]), float(row[yCol]), float(row[zCol]))*0.1)
+            formalCharges.append(row[formalChargeCol])
         positions = positions*unit.nanometers
 
         # Load the bonds.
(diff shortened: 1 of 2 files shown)


pdbfixer/pdbfixer.py
```diff
@@ -108,6 +110,98 @@ def __init__(self, topology, positions, terminal=None):
...
+        residueName = block.getObj('chem_comp').getValue("id")
+
+        descriptorsData = block.getObj("pdbx_chem_comp_descriptor")
+        typeCol = descriptorsData.getAttributeIndex("type")
+        smilesCol = descriptorsData.getAttributeIndex("descriptor")
+        smiles = None
+        for row in descriptorsData.getRowList():
+            if row[typeCol] in ["SMILES", "SMILES_CANONICAL"]:
+                smiles = row[smilesCol]
+                break
```
peastman, line 163: You record the SMILES string but never use it anywhere.  What is it for?

pdbfixer/pdbfixer.py
```diff
@@ -1448,20 +1567,34 @@ def _createForceField(self, newTopology, water):
             template = app.ForceField._TemplateData(resName)
             forcefield._templates[resName] = template
             indexInResidue = {}
+            # If we can't find formal charges in the CCD, make everything uncharged
+            formalCharges = defaultdict(lambda: 0)
```
peastman, line 1571: Or more simply `defaultdict(int)`?

pdbfixer/tests/test_charge_and_solvate.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    fixer = PDBFixer(pdbid=pdbCode)
+    fixer.findMissingResidues()
+
+    chains_to_cap = {chain for chain, resi in fixer.missingResidues}
+    for chainidx in chains_to_cap:
+        chain = [*fixer.topology.chains()][chainidx]
+        last_resi = len([*chain.residues()])
+        # Capping with GLY because ACE/NME currently breaks addMissingHydrogens
+        # Adding a cap keeps the protein compact and the addSolvent call quick
```
peastman, line 37: In what way does it break it?

pdbfixer/pdbfixer.py
```diff
@@ -354,6 +451,50 @@ def _initializeFromPDBx(self, file):
...
+        This method caches results in the ``PDBFixer`` object.
+
+        Parameters
+        ----------
+        residueName : str
+            The name of the residue, as specified in the PDB Chemical Component
+            Dictionary.
+        checkCache : bool
+            If ``False``, attempt to re-download the CCD entry regardless of
+            what is in the cache. Defaults to ``True``.
```
peastman, line 468: What's the reason for including this flag?  You never specify it when calling this method, and it's a private internal method not meant to be called by users, so the value is never False.

peastman (commented):
Sorry for the delay in getting back to you.  Your implementation looks good, and as you say, it's probably better than not doing it.

I'm still a little concerned about residues that can appear in multiple forms.  Take a look at the CCD definition of HIS, for example.  (I know it's a standard residue, and therefore this code won't be used for it.  I'm just using it as an example of what we can probably expect to find in nonstandard residues too.)  It can be positive, negative, or neutral depending on which hydrogens are present.  The CCD entry describes the positively charged, doubly protonated form.  But that isn't the most common form at pH 7.

On the positive side, if PDBFixer just added the hydrogens based on the CCD entry, that's the form it will use.  So this does ensure that if PDBFixer adds both hydrogens and solvent, they'll be consistent with each other.

The other thing I'm a bit unsure about is the handling of leaving atoms.  You require that no leaving atoms be present.  If any of them are, it won't match.  Would it be better to make it match whether or not the leaving atoms are present?

Yoshanuikabundi:
> Sorry for the delay in getting back to you. Your implementation looks good, and as you say, it's probably better than not doing it.

Yay!

> I'm still a little concerned about residues that can appear in multiple forms. Take a look at the CCD definition of HIS, for example. (I know it's a standard residue, and therefore this code won't be used for it. I'm just using it as an example of what we can probably expect to find in nonstandard residues too.) It can be positive, negative, or neutral depending on which hydrogens are present. The CCD entry describes the positively charged, doubly protonated form. But that isn't the most common form at pH 7.

An example of this is the CCD definition of CIT, citric acid. At pH 7, citric acid is most commonly triply deprotonated, but the CCD entry is neutral. I agree it would be nice to handle this case more carefully, both here and in the protonation code.

I can imagine some simple algorithm along the lines of "add 1 formal charge to any atom that has an extra hydrogen bonded to it relative to the CCD, and remove 1 formal charge from any atom that is missing a hydrogen" which might improve the behavior in this PR for "nonstandard" protonation states. This would have pathological behavior on a PDB file that is missing hydrogens though - carbanions aplenty! And if `_createForcefield(water=True)` is used for any process that commonly happens before protonation this could be very bad. So we'd want some further checks, like only applying this algorithm to certain elements or skipping it if the entire residue, molecule, chain, or file is missing hydrogens. But it's difficult to see how to differentiate in general between PDB files where a proton is missing and PDB files representing chemicals that lack that proton. For instance, we can contrive a case like fluoroformic acid, FCOOH, in which the deprotonated form FCOO- is indistinguishable in a OpenMM topology from the protonated form without hydrogens. We might be able to use the PDB charge column for this?

In general, I don't think this should hold up this PR, because the PR will only introduce regressions if the current behavior was correct by luck. I think detecting alternative forms is a substantial and orthogonal design problem.

> On the positive side, if PDBFixer just added the hydrogens based on the CCD entry, that's the form it will use. So this does ensure that if PDBFixer adds both hydrogens and solvent, they'll be consistent with each other.

This'll need to be factored in if the hydrogen adding system becomes pH aware for CCD sourced residues - the same mechanism will want to be applied here.

> The other thing I'm a bit unsure about is the handling of leaving atoms. You require that no leaving atoms be present. If any of them are, it won't match. Would it be better to make it match whether or not the leaving atoms are present?

Yeah I think I misunderstood what the point of leaving atoms were - they're for defining polymer linkages, but I thought they were about alternate protonation states. I'll fix this!

pdbfixer/pdbfixer.py
Yoshanuikabundi: I think I was using it in a previous iteration. I'll remove it.

pdbfixer/pdbfixer.py
Yoshanuikabundi: Nice, fixed

pdbfixer/tests/test_charge_and_solvate.py
Yoshanuikabundi: I was running into an issue where if I unconditionally capped the ends of a chain, chains that didn't have missing terminal loops would wind up with unprotonated caps. I mentioned it briefly in #299. I ended up attributing this to a bug in my own code and fixed it by only capping terminals with missing loops, and I've transferred that code over to this PR just now, but I can dig into it further if you like.

pdbfixer/pdbfixer.py
Yoshanuikabundi: This is another leftover from an earlier iteration and I've removed it.

peastman:
The test case is failing:

```
>       assert soluteCharge == numCl - numNa
E       assert -21 == (45 - 67)
```

peastman:
I think you're right: this won't always get the right answer, but it's more likely to get the right answer than the current code, so it's still an improvement.

I think it would be worth also allowing the user to pass a ForceField to `addSolvent()`, which would be used instead of creating one based on templates.  It's hardly any extra code, and it would allow users with custom residues to make sure they're getting the right charge based on the actual force field they'll be using to simulate it.  If you prefer though, we can do that in a separate PR.

peastman:
Just checking in on this so it doesn't get forgotten.  The test case is still failing.

Yoshanuikabundi:
Sorry Peter, I've been distracted! The failing test was a timeout from code I didn't touch downloading PDB files from the PDB. Hopefully it'll succeed this time - I think this one's good to go.

peastman:
Just checking in again. :)  See the one comment above, where if it can't download one definition, it won't return any others either.  I think that line needs to be changed?

pdbfixer/pdbfixer.py
```diff
@@ -1247,33 +1359,13 @@ def _downloadNonstandardDefinitions(self):
...
-
-                try:
-                    file = urlopen(f'https://files.rcsb.org/ligands/download/{name}.cif')
-                    contents = file.read().decode('utf-8')
-                    file.close()
-                except:
-                    continue
+                ccdDefinition = self._downloadCCDDefinition(name)
+                if ccdDefinition is None:
+                    return None
```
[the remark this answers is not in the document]
Yoshanuikabundi: Yeah I think that makes more sense with how I ended up implementing `_createForceField()`. My reasoning for `return None` was to be conservative - if any residue in the chain isn't in the CCD, then the assumptions of the method are violated and the entire chain cannot be charged, so there's no point in wasting the user's time downloading definitions for the rest of the chain. But `_createForceField()` works on a residue-by-residue basis, so an unknown residue won't interrupt anything there.

```suggestion
                    continue
```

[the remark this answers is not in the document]
Yoshanuikabundi: Oh yeah and that goes double for the fact that this is shared between all the methods that download residue definitions! Whoops.

Yoshanuikabundi:
Sorry I've been off sick - I agree, and have made that change. Thanks for being so patient!

peastman:
Thanks, that looks good.  I'm sorry you've been sick!

PR #303, opened 2024-08-08
Added optimizeHydrogenBonds()

peastman:
This adds a function for optimizing the hydrogen bonding network of a protein.  It does it in two ways:

- Rotating the terminal groups of HIS, ASN, and GLN sidechains by 180 degrees.
- Switching HIS residues between the HID and HIE forms.

It considers all possible combinations to find the one that maximizes the number of hydrogen bonds.

### DIFF
--- a/pdbfixer/hbonds.py
+++ b/pdbfixer/hbonds.py
@@ -9,6 +9,7 @@ from collections.abc import Iterator
 from typing import List
 
 class HydrogenBondOptimizer(object):
+    """This class rotates sidechains to optimize the hydrogen bond network."""
     
     def __init__(self, fixer):
         topology, positions = self.protonateHistidines(fixer)
@@ -30,7 +31,7 @@ class HydrogenBondOptimizer(object):
             self.residueDonors[residue] = []
             self.residueAcceptors[residue] = []
             for atom in residue.atoms():
-                if atom.element == hydrogen and parent[atom].element in (oxygen, nitrogen, fluorine):
+                if atom.element == hydrogen and atom in parent and parent[atom].element in (oxygen, nitrogen, fluorine):
                     self.residueDonors[residue].append((atom, parent[atom]))
                 elif atom.element in (oxygen, nitrogen, fluorine):
                     self.residueAcceptors[residue].append(atom)
@@ -162,7 +163,7 @@ class HydrogenBondOptimizer(object):
 
         # Create the final topology and positions.
 
-        fixer.topology, fixer.positions = self.deleteExtraHydrogens(fixer, topology, positions)
+        fixer.topology, fixer.positions = self.deleteExtraHydrogens(fixer, topology, newPositions)
 
     def protonateHistidines(self, fixer):
         """Add a second hydrogen to neutral histidines so we can identify bonds formed by both of them."""
@@ -256,13 +257,13 @@ class HydrogenBondOptimizer(object):
 
     def isHbond(self, d: mm.Vec3, h: mm.Vec3, a: mm.Vec3):
         """Decide whether three atoms could form a hydrogen bond."""
-        if unit.norm(d-a) > 0.35:
+        if unit.norm(h-a) > 0.25:
             return False
         deltaDH = h-d
         deltaHA = a-h
         deltaDH /= unit.norm(deltaDH)
         deltaHA /= unit.norm(deltaHA)
-        return np.arccos(np.dot(deltaDH, deltaHA)) < 50*np.pi/180
+        return np.arccos(np.dot(deltaDH, deltaHA)) < 60*np.pi/180
 
     def countHbonds(self, res1: app.Residue, res2: app.Residue, pos1: List[mm.Vec3], pos2: List[mm.Vec3]):
         """Count the number of hydrogen bonds between two residues."""
--- a/pdbfixer/hbonds.py
+++ b/pdbfixer/hbonds.py
@@ -1,0 +1,292 @@
+import openmm as mm
+import openmm.app as app
+import openmm.unit as unit
+from openmm.app.element import hydrogen, oxygen, nitrogen, fluorine
+import numpy as np
+import itertools
+from collections import defaultdict
+from collections.abc import Iterator
+from typing import List
+
+class HydrogenBondOptimizer(object):
+    
+    def __init__(self, fixer):
+        topology, positions = self.protonateHistidines(fixer)
+
+        # Record the parent atom that every hydrogen is bonded to.
+
+        parent = {}
+        for atom1, atom2 in topology.bonds():
+            if atom1.element == hydrogen and atom2.element != hydrogen:
+                parent[atom1] = atom2
+            elif atom2.element == hydrogen and atom1.element != hydrogen:
+                parent[atom2] = atom1
+
+        # Record the donors and acceptors in each residue.
+
+        self.residueDonors = {}
+        self.residueAcceptors = {}
+        for residue in topology.residues():
+            self.residueDonors[residue] = []
+            self.residueAcceptors[residue] = []
+            for atom in residue.atoms():
+                if atom.element == hydrogen and parent[atom].element in (oxygen, nitrogen, fluorine):
+                    self.residueDonors[residue].append((atom, parent[atom]))
+                elif atom.element in (oxygen, nitrogen, fluorine):
+                    self.residueAcceptors[residue].append(atom)
+
+        # Record the groups of atoms we will try to rotate.
+
+        rotations = {
+            'ASN': (('CB', 'CG'), ('OD1', 'ND2', 'HD21', 'HD22')),
+            'GLN': (('CG', 'CD'), ('OE1', 'NE2', 'HE21', 'HE22')),
+            'HIS': (('CB', 'CG'), ('ND1', 'CD2', 'CE1', 'NE2', 'HD1', 'HD2', 'HE1', 'HE2'))
+        }
+        residueRotations = {}
+        for residue in topology.residues():
+            if residue.name in rotations:
+                axis, atoms = rotations[residue.name]
+                residueRotations[residue] = ([atom for atom in residue.atoms() if atom.name in axis],
+                                             [atom for atom in residue.atoms() if atom.name in atoms])
+
+        # For each residue we will be modifying, record all other residues that are close enough
+        # to potentially form hydrogen bonds.
+
+        self.residueNeighbors = {}
+        residueBounds = dict((residue, BoundingSphere(residue.atoms(), positions)) for residue in topology.residues())
+        for res1 in residueRotations:
+            self.residueNeighbors[res1] = []
+            bounds = BoundingSphere(residueRotations[res1][1], positions)
+            for res2 in topology.residues():
+                if res1 != res2 and bounds.distance(residueBounds[res2]) < 0.5:
+                    close = False
+                    for atom1 in residueRotations[res1][1]:
+                        for atom2 in res2.atoms():
+                            delta = positions[atom1.index]-positions[atom2.index]
+                            if unit.norm(delta) < 0.5:
+                                close = True
+                                break
+                        if close:
+                            break
+                    if close:
+                        self.residueNeighbors[res1].append(res2)
+
+        # Divide them into clusters that need to be analyzed together.
+
+        clusters = [[res] for res in self.residueNeighbors]
+        converged = False
+        while not converged:
+            converged = True
+            for i in range(len(clusters)):
+                if i >= len(clusters):
+                    break
+                c1 = clusters[i]
+                for j in range(i+1, len(clusters)):
+                    c2 = clusters[j]
+                    if any(res2 in self.residueNeighbors[res1] for res1, res2 in itertools.product(c1, c2)):
+                        c1 += c2
+                        del clusters[j]
+                        converged = False
+                        break
+
+        # Create the rotated version of the coordinates.
+
+        rotatedPositions = positions[:]
+        for residue in residueRotations:
+            axis, atoms = residueRotations[residue]
+            center = positions[axis[1].index]
+            e = center-positions[axis[0].index]
+            e /= unit.norm(e)
+            for atom in atoms:
+                d = positions[atom.index]-center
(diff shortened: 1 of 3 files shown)


pdbfixer/hbonds.py
```diff
@@ -0,0 +1,293 @@
...
+
+        # Record the donors and acceptors in each residue.
+
+        self.residueDonors = {}
+        self.residueAcceptors = {}
+        for residue in topology.residues():
+            self.residueDonors[residue] = []
+            self.residueAcceptors[residue] = []
+            for atom in residue.atoms():
+                if atom.element == hydrogen and atom in parent and parent[atom].element in (oxygen, nitrogen, fluorine):
```
alex-sbaq, line 34: We might need to include sulfur to capture Cysteine sidechains. Do we need fluorine in this list? H-F bonds are unlikely.

pdbfixer/hbonds.py
```diff
@@ -0,0 +1,293 @@
...
+                        bonds += residueRotationHbondChange[residue]
+                    for neighbor in self.residueNeighbors[residue]:
+                        if (residue, neighbor) in pairRotationHbonds:
+                            neighborRotated = combination[cluster.index(neighbor)]
+                            bonds += pairRotationHbonds[(residue, neighbor)][rotated][neighborRotated]
+                if bonds > bestBonds:
+                    bestCombination = combination
+                    bestBonds = bonds
+
+            # Set the positions to the optimal set of rotations.
```
alex-sbaq, line 157: Do we check that rotations do not cause clashes? The attached example is from PDB code 5UEX (H437 and V439). The histidine sidechain rotates to put two protons in close proximity. But, this code would probably capture this as two H-bonds and favor it (histidine N-H > backbone N and backbone N-H > histidine N). I would expect the backbone N-H to hydrogen bond with the histidine sidechain.
prot_clash

peastman: It doesn't check for clashes, because most of them will be resolved by energy minimization.  In the image above, I don't see what the clash is?  The hydrogen pointing off to the left looks like it's out of the way of the other atoms.  It wouldn't be perceived as a second hydrogen bond because it isn't in line.  The angle between the D-H and H-A vectors can't be more than 60 degrees.  Also, the H-A distance has to be under 2.5 Å, and in that image it's 3.1 Å.

pdbfixer/hbonds.py
peastman: Fluorine is one of the most common hydrogen bond acceptors.  It's very common in drug molecules.

alex-sbaq: Sorry, the fluorine comment was meant for the donor list. Definitely keep it for the acceptors 😄

peastman: According to [Wikipedia](https://en.wikipedia.org/wiki/Hydrogen_bond), F is also a very common donor.

pdbfixer/hbonds.py
alex-sbaq: Added a bit more context to the image this time. The backbone N-H is distorted (120 degrees). When minimized to ~180 degrees, it will be placed directly in line with the histidine sidechain N-H. Also when minimized, it could form a hydrogen bond to the histidine (if it was in the other protonation state).
prot_clash2

pdbfixer/hbonds.py
alex-sbaq: Yes, but to be a donor, F needs to be bound to hydrogen. F can only make one bond, so it would be an isolated H-F molecule.

pdbfixer/hbonds.py
peastman: I don't think I understand what you're saying.  If it rotates by 120 degrees as shown in the image, it will be pointing directly away from the HIS.  That will neither clash nor form a hydrogen bond.

What exactly is the issue you're describing, not just in this case but in general?  And what are you suggesting it should do differently?

pdbfixer/hbonds.py
```diff
@@ -0,0 +1,293 @@
...
+                            newTopology.addBond(newAtom, atomsByName['NE2'])
+                            axis = posByName['NE2'] - 0.5*(posByName['CD2']-posByName['CE1'])
+                            axis /= unit.norm(axis)
+                            newPositions.append(posByName['NE2'] + axis)
+                        elif 'HD1' not in atomsByName and 'HE2' in atomsByName:
+                            newAtom = newTopology.addAtom('HD1', hydrogen, newResidue)
+                            newTopology.addBond(newAtom, atomsByName['ND1'])
+                            axis = posByName['ND1'] - 0.5*(posByName['CG']-posByName['CE1'])
+                            axis /= unit.norm(axis)
+                            newPositions.append(posByName['ND1'] + axis)
```
alex-sbaq, line 198: I think there is a sign slip and a missing bond length in this line, and between them they end up setting the tautomer choice further down.

`posByName['CD2']-posByName['CE1']` is a ring edge vector rather than a bisector, so `axis` is `NE2` minus a small vector, which is still an absolute position rather than a direction. Normalising that gives a unit vector pointing from the coordinate origin toward the residue, and nothing scales it to a bond length, so the new hydrogen lands 1.0 nm from NE2.

On 1BHL every hydrogen added here sits 10.00 Å from its nitrogen, against 1.18 Å for the ones pdbfixer already placed. Translating the whole structure by 5 nm swings the placement direction by 28 to 47 degrees, so the result tracks where the molecule happens to sit in the file.

That feeds into `deleteExtraHydrogens`, which decides which ring hydrogen to keep by asking which one makes a hydrogen bond. Most of the time the placeholder finds nothing and gets deleted as intended. When it happens to land near something, `isHbond` accepts it and the other hydrogen, the one pdbfixer placed, is the one removed. On 4HHB, B143 and C103 come out with HE2 sitting 10 Å from NE2 on every prep I ran, and D97 on about two thirds of them. 1TIM (A95, A115) and 1M17 (A781, A811) do the same. 1BHL and 3EIY hit it less often, roughly a quarter and a third of preps, the variation coming from `addMissingAtoms` being stochastic. The acceptors the placeholder reaches are 10 to 12 Å from the histidine's own ring, one of them on a different chain, which is about what you would expect for something a nanometre out.

Across the five structures, every tautomer change I saw traced back to this rather than to the hydrogen bonding, so at the moment the placement looks like it is doing the deciding. With the fix below, 1BHL switches 3 of its 6 histidines and the long bonds go away.

Worth saying the geometry recovers on its own. amber14 minimisation pulls the bond back to 1.01 Å without trouble, including with `constraints=HBonds`, so nobody's simulation is at risk here. The part that persists is the tautomer, since minimisation cannot change one. On 4HHB the current code and the patched version land on different tautomers for 7 of 38 histidines. Two of those are the cases above where the placed hydrogen was removed. The other five are quieter: the placeholder found nothing, the code kept the input tautomer, and there is no geometric trace left afterwards to show it happened.

The same two things are in the ND1 branch, so the suggestion covers both.

```suggestion
                            axis = posByName['NE2'] - 0.5*(posByName['CD2']+posByName['CE1'])
                            axis /= unit.norm(axis)
                            newPositions.append(posByName['NE2'] + 0.101*axis)
                        elif 'HD1' not in atomsByName and 'HE2' in atomsByName:
                            newAtom = newTopology.addAtom('HD1', hydrogen, newResidue)
                            newTopology.addBond(newAtom, atomsByName['ND1'])
                            axis = posByName['ND1'] - 0.5*(posByName['CG']+posByName['CE1'])
                            axis /= unit.norm(axis)
                            newPositions.append(posByName['ND1'] + 0.101*axis)
```

pdbfixer/hbonds.py
```diff
@@ -0,0 +1,293 @@
...
+            # Loop over all possible combinations of which residues to rotate.
+
+            for combination in itertools.product(*[[0,1]]*len(cluster)):
+                bonds = 0
+                for rotated, residue in zip(combination, cluster):
+                    if rotated == 1:
+                        bonds += residueRotationHbondChange[residue]
+                    for neighbor in self.residueNeighbors[residue]:
+                        if (residue, neighbor) in pairRotationHbonds:
+                            neighborRotated = combination[cluster.index(neighbor)]
```
alex-sbaq, line 151: This one raises on structures where the neighbour relation comes out one-directional.

`residueNeighbors` is built at 56 to 73 by testing only the rotating atoms of res1 against all atoms of res2, so `res2 in N(res1)` does not imply the reverse. When that happens between two rotatable residues, the pair key at 124 can exist while the two residues land in different clusters, and `cluster.index(neighbor)` throws.

1BRS hits it on every prep I tried:

```
ValueError: <Residue 358 (ASN) of chain 3> is not in list
```

6LU7 hits it on 2 preps out of 3 and 2SRC on 1 out of 3, the variation again coming from `addMissingAtoms` rebuilding sidechains differently each time.

Symmetrising `residueNeighbors` just after the loop ends at line 73 covers it, and it also picks up a quieter version of the same thing: when the relation is one-directional and the indices run the wrong way, the interaction between two rotatable residues drops out of the objective entirely, since line 112 skips it for being rotatable and line 123 skips it on the index test.

pdbfixer/pdbfixer.py
```diff
@@ -1340,6 +1341,18 @@ def _describeVariant(self, residue, definitions):
...
+        """Optimize the hydrogen bond network.
+
+        This should be called after hydrogens have been added.  It considers two types of changes:
+
+        1. 180 degree rotations of the terminal groups in HIS, ASN, and GLN sidechains.
+        2. For neutral HIS residues, exchanging the hydrogen position between ND1 and NE2.
+
+        It considers all possible combinations of them to find the one producing the largest number of hydrogen bonds.
+        """
+        HydrogenBondOptimizer(self)
```
alex-sbaq, line 1354: If this gets called before hydrogens exist, it returns without doing anything and without raising. With no hydrogens there are no donors, every pair scores zero, the all-unflipped combination wins, and `deleteExtraHydrogens` has nothing to delete. On 1BHL I get 1101 atoms in, 1101 atoms out, positions bitwise identical and no exception. The docstring does say it should be called after hydrogens are added, so this is a misuse, but the way it surfaces is a pipeline that reports success and quietly skips the step. `hydrogen` is already imported at line 40 if a guard seems worth it.

```suggestion
        if not any(atom.element == hydrogen for atom in self.topology.atoms()):
            raise ValueError('optimizeHydrogenBonds() requires a structure that already has hydrogens.  Call addMissingHydrogens() first.')
        HydrogenBondOptimizer(self)
```

pdbfixer/hbonds.py
alex-sbaq: Coming back to this with something more concrete than the fluorine question.

The part I keep coming back to on this line is the acceptor side. `atom.element in (oxygen, nitrogen, fluorine)` puts every nitrogen in the acceptor list, and a fair number of them cannot accept. In 3EIY that is 211 of 598, so 35 percent: 162 backbone amide N, 13 LYS NZ, 12 PRO N, 12 arginine guanidinium N, plus TRP NE1 and the ASN ND2 and GLN NE2 amide nitrogens.

Those last two are the ones I would worry about most here. The reason to flip an ASN or GLN at all is that the crystallographic O and N assignment is ambiguous and hydrogen bond complementarity is what resolves it. If ND2 and NE2 also count as acceptors, the two orientations score closer together and the flip has less to work with.

The geometry cut in `isHbond` does screen some of these out, though incidentally rather than as a matter of chemistry, and it does not clear any category. It takes the non-accepting share from roughly half of the pairs tested down to about a sixth of the ones accepted, and every structure I looked at still scores somewhere between a dozen and thirty hydrogen bonds onto an atom that cannot accept one, at a median of about 1.9 Å and 145 degrees. Those sit inside the window rather than at its edge. The lists are also not narrowed after they are built, so whatever goes in at line 36 is what gets scored.

It does move the answer in at least one case I can point at. Details are on the other thread, but on 3EIY this leads the code to flip HIS111 into a clash with Lys113 and trade away a real hydrogen bond to a water in the process.

A residue and atom name table is the usual way to handle this, along the lines of what Reduce and HBPLUS do. Sulfur would be worth picking up at the same time, for CYS SG and MET SD.

pdbfixer/hbonds.py
alex-sbaq: Apologies for the very long silence here, and my 5UEX example was not a good one. Let me try the general question again, with a case that reproduces.

The general version of the concern is that the objective counts attractive terms and has no repulsive one, so a rotation that gains a bond while creating an overlap scores the same as one that gains a bond cleanly. You are right that minimisation cleans up geometry, and it does. What it cannot do is reverse the rotation, because the flip is a discrete state. Minimisation relieves the overlap by distorting locally, so the flip stays.

There is a sharper version, which is what I should have said the first time: because acceptors are typed by element, some contacts that are really clashes get counted as hydrogen bonds rather than penalised.

3EIY, HIS111 and Lys113. Standard pipeline, then `optimizeHydrogenBonds()`. Before the flip, ND1-H points at HOH240 at 3.32 Å. After the flip, over three replicates:

```
HD1 to Lys113 N            1.76 to 1.95 A     isHbond accepts, it needs <= 2.50
ND1-HD1 to N angle          130 to 151 deg    isHbond accepts, it needs >= 120
HD1 to Lys113 amide H      1.87 to 1.95 A     H to H, vdW contact is about 2.4
ND1 to nearest water              4.11 A      was 3.32 before the flip
```

Lys113 N is a backbone amide nitrogen, so it cannot accept, and it already carries its own hydrogen. That contact scores as a hydrogen bond, the flip wins on count, and the water hydrogen bond goes. The ND1 displacement is 2.34 Å every time. Skip `addMissingAtoms` if you want it fully deterministic, since HIS111 and Lys113 are both complete in the crystal.

This one is easy to picture, but it is not the only one. Counting across whole structures, every one I checked scores a dozen to thirty hydrogen bonds per run onto nitrogens that cannot accept one, with typical geometry inside the acceptance window rather than at its edge. I have put the breakdown on the line 34 thread.

On what to do differently, I think the acceptor typing is most of it. Excluding the nitrogens that cannot accept, HIS111 stays put and the contact goes back to about 6.2 Å. A lower bound on the H to acceptor distance would pick up some of the remainder, but the typing looks like the load-bearing part.

PR #308, opened 2024-09-30
Soft core force field includes torsions with wildcards

### LINKED ISSUE #307: PDBFixer - in some cases added residues have amides with non-trans OCNH stereochemistry
Hi There,

I first want to say that for my cases PDBFixer is superior to other methods. For the case of 7zaw.pdb - Human GPC3 having missing atoms and residues (many) - PDBFixer is able to add the missing residues (that can be very long) extremely well, while other methods just aren't anywhere near as good. I noticed in some cases that linked residues can have amides with non-trans OCNH stereochemistry, and then MM minimization can actually make cis OCNH structures. 

The test example I provide is 7zaw.pdb (RCSB) where PDBFixer produces the structure in 7zaw_PDBFixer. Examples of residue links having non-trans OCNH structures include:

3(G)-4(D)
5(A)-6(T)
21(L)-22(K)
28(P)-29(V)
161(D)-162(S)
162(S)-163(A)
284(R)-285(I)

I was wondering if there is a way to add residues whereby trans OCNH stereochemistry is enforced. I understand that the residues may have unfavorable positions but I'm hoping that a MM minimization will 'repair' them, while also keeping amide trans OCNH structures.

It's possible I'm not using PDBFixer correctly and thus I end up with non-trans OCNH amides, if so please let me know as this would be the easiest solution. 

Thank you for any help you can offer, I very much appreciate it.

Joe

P.S. Note that I had to add ".txt" to end of pdb files so that github would upload them.

[7zaw.pdb.txt](https://github.com/user-attachments/files/17192510/7zaw.pdb.txt)
[7zaw_PDBFixer.pdb.txt](https://github.com/user-attachments/files/17192511/7zaw_PDBFixer.pdb.txt)

peastman:
Fixes #307.

The soft-core force field used for energy minimization was created based on Amber ff99SB-ILDN.  The script to create it didn't handle wildcards correctly.  That meant the force field was missing any torsion whose definition depended on a wildcard.

### DIFF
--- a/devtools/createSoftForcefield.py
+++ b/devtools/createSoftForcefield.py
@@ -51,18 +51,20 @@ print(' <AtomTypes>')
 omitTypes = set()
 omitClasses = set()
 for atomType in forcefield._atomTypes:
-    (atomClass, mass, element) = forcefield._atomTypes[atomType]
-    if element is None or element == elem.hydrogen:
+    data = forcefield._atomTypes[atomType]
+    if data.element is None or data.element == elem.hydrogen:
         omitTypes.add(atomType)
-        omitClasses.add(atomClass)
+        omitClasses.add(data.atomClass)
     else:
-        print('  <Type name="%s" class="%s" element="%s" mass="%g"/>' % (atomType, atomClass, element.symbol, mass))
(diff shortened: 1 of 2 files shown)


peastman:
It seems this doesn't really do much for #307, but it's still a legitimate fix.  It means we do the energy minimization with a slightly better force field.

PR #322, opened 2025-01-06
Fix exception with unknown residues

### LINKED ISSUE #321: PDBFixer fails to solvate due to a None return, without a handled exception
I'm working on a use case I've run for many years for a course I teach, where I take a (probably somewhat nonstandard) PDB file containing a host-guest complex and solvate with PDBFixer. Apparently with recent updates to PDBFixer my example has broken.

Previously, the following worked:
```
fixer = PDBFixer( 'datafiles/complex.pdb')
fixer.addSolvent(padding=unit.Quantity( 1.0, unit.nanometers), ionicStrength=unit.Quantity( 20, unit.millimolar))
```

now it fails with this (final) error, resulting from the line ` typeCol = descriptorsData.getAttributeIndex("type")`:
```
AttributeError: 'NoneType' object has no attribute 'getAttributeIndex'
```

Looking at the full traceback it looks like a lookup of the ccD residue definition/formal charges may have returned None which results in no type being available here, hence the failed AttributeError, but I'm not totally sure. 

I would guess that the problem here is either (a) un-handled default/fallback behavior when a lookup fails, or (b) exception handling to give me a more useful error message about what is wrong with my PDB file (which may well be noncompliant). As noted, this is code which previously worked so changes in PDBfixer in the last 8-10 months have broken it. 

Suggestions?

[PDB file available here](https://www.dropbox.com/scl/fi/8ke0ijr9wb52hru80eora/complex.pdb?rlkey=dha233xz5pi19iwqm6fk37yws&dl=0)

peastman:
Fixes #321.  The error happened in a file containing the residue ID UNL (unknown ligand).  The CCD contains an entry for it, but it's a minimal file with no atoms.  That led to an exception.

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -150,10 +150,10 @@ class CCDResidueDefinition:
 
         residueName = block.getObj('chem_comp').getValue("id")
 
-        descriptorsData = block.getObj("pdbx_chem_comp_descriptor")
-        typeCol = descriptorsData.getAttributeIndex("type")
-
         atomData = block.getObj('chem_comp_atom')
+        if atomData is None:
+            # The file doesn't contain any atoms, so report that no definition is available.
+            return None
         atomNameCol = atomData.getAttributeIndex('atom_id')
         symbolCol = atomData.getAttributeIndex('type_symbol')
         leavingCol = atomData.getAttributeIndex('pdbx_leaving_atom_flag')


davidlmobley:
Thanks. How soon could we expect this to be in a release? (My course conda/mamba installs openmm, so I need to know if this will be in a release before I get to the lecture, which is probably mid-February.)

peastman:
I think we can do that.  Let's wait a little longer to give time for more bugs to come up.  If I haven't created a new release by the end of the month, remind me and I'll do it then.

davidlmobley:
You asked me to ping you if a release wasn't out by the end of January; I'm not seeing one yet. Can we get this into a release?

peastman:
I'll see if I can get it out today.

peastman:
The packages for 1.11 are up.

davidlmobley:
Thank you!

PR #338, opened 2025-09-22
Add legacy-cgi to replace cgi

### LINKED ISSUE #318: Remove use of cgi module
The PDBFixer GUI uses the cgi module, which was removed in python 3.13.  There are instructions on how to replace it at https://peps.python.org/pep-0594/#cgi.

epretti:
Adds the legacy-cgi package to replace the deprecated cgi module that is gone from Python 3.13.  Resolves #318.

Does `devtools/conda-recipe/meta.yaml` need to be updated too?  This looks old and outdated compared to the [`meta.yaml` in conda-forge/pdbfixer-feedstock](https://github.com/conda-forge/pdbfixer-feedstock/blob/main/recipe/meta.yaml) so I assume it's not being used currently?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.11", "3.12", "3.13"]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


peastman:
We should only add it to `install_requires` when installing on Python 3.13 or later.

Yes, we'll need to update the conda recipe as well.  I'll be doing an updated release soon, so I'll make the change as part of building the release.

peastman:
Oh, sorry, you mean the one in this repo.  Correct, that's obsolete.

epretti:
Are `.travis.yml` and `devtools/travis-ci` also obsolete?  If so, I could remove them (probably in a separate PR) with the `devtools/conda-recipe` that is in this repository.

PR #346, opened 2026-03-05
Improve ACE/NME cap geometry with NeRF-based placement

aqemia-richard-gildea:
## Problem

`_addMissingResiduesToChain` uses Kabsch SVD overlay to place all missing residues, including terminal caps (ACE/NME). This produces distorted geometry for caps:

- **ACE** has only 2 matching backbone atoms (C, O) — rotation around the C-O axis is underdetermined
- **NME** matches 2 atoms by name (N, C) but the match is semantically wrong: NME's sp3 methyl "C" aligns to the adjacent residue's sp2 backbone "C", which are at different positions relative to N
- Neither cap has CA, so the trans-peptide rotation step (line 794) is skipped
- Result: bond angles and omega dihedrals can be significantly off, depending on the structure

## Solution

Use the **NeRF algorithm** (Natural Extension Reference Frame, [Parsons et al. 2005](https://doi.org/10.1002/jcc.20237)) to place cap atoms from internal coordinates (bond length, bond angle, dihedral) relative to three backbone reference atoms (N, CA, C) of the adjacent residue. This is fully determined regardless of the number of template atoms and produces ideal geometry in one shot.

### Changes

**`pdbfixer/pdbfixer.py`**:
- Add `_placeAtomNeRF()`: general NeRF atom placement from 3 reference points + internal coordinates
- Add `_buildAceCap()` / `_buildNmeCap()`: build cap positions using bond parameters from [Engh & Huber (1991)](https://doi.org/10.1107/S0108767391001071), Acta Cryst. A47, 392-400 ([tabulated values](https://www.ebi.ac.uk/thornton-srv/software/PROCHECK/manual/manappa.html))
- Modify `_addMissingResiduesToChain`: route ACE/NME to NeRF placement when adjacent residue has full backbone (N, CA, C); fall back to existing Kabsch overlay otherwise

**`pdbfixer/tests/test_nerf_caps.py`** (new):
- Unit tests for NeRF placement (bond length, angle, collinear guard)
- Bond length and omega dihedral tests for ACE/NME caps, parametrized over identity and rotated reference frames
- End-to-end integration test verified against AMBER14 force field

### Test results

All 34 tests pass (12 new + 22 existing).

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -275,6 +275,61 @@ def _dihedralRotation(points, angle):
         [axis[2]*axis[0]*(1-ct) - axis[1]*st,    axis[2]*axis[1]*(1-ct) + axis[0]*st,    axis[2]*axis[2]*(1-ct) + ct        ]
     ])
 
+def _placeAtomNeRF(a1, a2, a3, bond_length, bond_angle, dihedral):
+    """Place atom X bonded to a3 using the NeRF algorithm (Parsons et al. 2005).
+
+    Given three reference points a1, a2, a3, compute the position of X such that
+    dist(a3, X) = bond_length, angle(a2, a3, X) = bond_angle, and
+    dihedral(a1, a2, a3, X) = dihedral.
+
+    All lengths in nanometers, angles in radians.
+    """
+    bc = a3 - a2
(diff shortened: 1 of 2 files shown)


peastman:
Just checking in on this again.  Can you answer the above question?

PR #347, opened 2026-03-06
Fix position of added OXT atom

aqemia-thomas-holder:
**Problem description:** The computation of the OXT atom position is wrong, it's flipped to the wrong side. Energy minimization often fixes the position, but not always. A good example is [4JSV](https://github.com/Aqemia/pdbfixer/blob/master/pdbfixer/tests/data/4JSV.pdb).

**Summary of changes:**

- Fixed inverted sign: `+2*v` → `-2*v`
- Added pipeline test

**Example before and after:**

```sh
pdbfixer pdbfixer/tests/data/4JSV.pdb --output=oxtfix-4JSV.pdb --add-atoms=heavy
```
image

### DIFF
--- a/pdbfixer/pdbfixer.py
+++ b/pdbfixer/pdbfixer.py
@@ -289,6 +289,14 @@ def _findUnoccupiedDirection(point, positions):
     direction /= unit.norm(direction)
     return direction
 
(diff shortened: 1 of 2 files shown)


peastman (commented):
Thanks!  I really appreciate the fix.  I have a few suggestions for making the code clearer.

peastman:
Looks good, thanks!

PR #349, opened 2026-06-26
Modified

TopGyro:
Changed the setup.cfg and setup.py as to include fix that would allow for pdbfixer installation to be easier.


peastman:
This change is incorrect.  It really does require 8.2.

Issue #350, opened 2026-07-29
Gaps are not filled after a residue is mutated


dww100:
As the title says - if you use applyMutations then you don't get gaps filled in your chain.

[Apologies for the very AI rest of the content - found while debugging something else and though worth flagging]

**Example**
Six residues in SEQRES, five in the ATOM records (ASP 3 is missing), and the mutation touches an unrelated residue.

```
from io import StringIO
from pdbfixer import PDBFixer

PDB = """\
SEQRES   1 A    6  ALA VAL ASP GLU PHE LYS
ATOM      1  N   ALA A   1      -1.200   0.000   0.000  1.00 20.00           N
ATOM      2  CA  ALA A   1       0.000   0.000   0.000  1.00 20.00           C
ATOM      3  C   ALA A   1       1.200   0.400   0.000  1.00 20.00           C
ATOM      4  O   ALA A   1       1.400   1.600   0.000  1.00 20.00           O
ATOM      5  N   VAL A   2       2.600   0.000   0.000  1.00 20.00           N
ATOM      6  CA  VAL A   2       3.800   0.000   0.000  1.00 20.00           C
ATOM      7  C   VAL A   2       5.000   0.400   0.000  1.00 20.00           C
ATOM      8  O   VAL A   2       5.200   1.600   0.000  1.00 20.00           O
ATOM      9  N   GLU A   4      10.200   0.000   0.000  1.00 20.00           N
ATOM     10  CA  GLU A   4      11.400   0.000   0.000  1.00 20.00           C
ATOM     11  C   GLU A   4      12.600   0.400   0.000  1.00 20.00           C
ATOM     12  O   GLU A   4      12.800   1.600   0.000  1.00 20.00           O
ATOM     13  N   PHE A   5      14.000   0.000   0.000  1.00 20.00           N
ATOM     14  CA  PHE A   5      15.200   0.000   0.000  1.00 20.00           C
ATOM     15  C   PHE A   5      16.400   0.400   0.000  1.00 20.00           C
ATOM     16  O   PHE A   5      16.600   1.600   0.000  1.00 20.00           O
ATOM     17  N   LYS A   6      17.800   0.000   0.000  1.00 20.00           N
ATOM     18  CA  LYS A   6      19.000   0.000   0.000  1.00 20.00           C
ATOM     19  C   LYS A   6      20.200   0.400   0.000  1.00 20.00           C
[...]

PR #351, opened 2026-07-29
Align residues to SEQRES instead of laying them out by residue number

### LINKED ISSUE #154: Fail to find missing residues on PDB with insertion code
The following code:
```python
from pdbfixer import PDBFixer
p = PDBFixer(pdbid='1a0q')
p.findMissingResidues()
```
find no missing residues although the chain `L` has 1 and chain `H` has 12.

This is due to residue numbering that reflects insertions and deletions (from another species I guess), which makes https://github.com/pandegroup/pdbfixer/blob/master/pdbfixer/pdbfixer.py#L611 never match any sequence.

I think that residue number cannot be use to find missing residues in the structure, even if taking into account insertion codes. This could be solved by aligning the residues with the sequence from SEQRES or using `_pdbx_poly_seq_scheme.ndb_seq_num` and `_pdbx_poly_seq_scheme.pdb_seq_num` from mmCIF or parsing `REMARK 465` from PDB, but I don't know how reliable this data is?

## Issues in chain L

There is a gap in residue numbering (residue number 95 is missing) which does not reflect a missing residue in the structure.

## Issues in chain H

There is a bunch of insertions (52A, 82A, 82B, 82C, 100A and 100B) that makes different residues have the same residue number.

### LINKED ISSUE #170: More rigorous implementation of findMissingResidues needed (e.g., doesn't work for antibodies)

### LINKED ISSUE #219: PDBs from RCSB failed to reconstruct
Hi,

I have been testing PDBFixer for some time and it works great in most cases. Unfortunately, I have also found that there are several PDBs for which the missing residues are not reconstructed (6RPA, 6RPB and 1AO7). PDBFixer just does not find the missing parts in the bakcbone and thus it does not reconstruct them. I have tried several approaches: first I checked if the SEQRES is well defined (even created own proper SEQRES), which was true. Then, I realized that the amino acids in these PDBs are not numbered in order for the continuous parts (not missing ones) - I also fixed that. Another fix was also to remove letters for the variant mutations done in the experiments (in comparison to reference sequence). 
Unfortunately, none of the above worked and at the moment there is effort to check through the code and find the reason. But maybe I could find help from you. All the suggestions and help will be very appreciated!

All the best,
    Sławek

dww100:
Fixes #219. Fixes #170. Fixes #154.

These three issues have the same root cause, so one change closes all of them.

For transparency all code here was written by Claude but at first glance seemed
close enough to what I was thinking of doing and tests seem sensible. The 
wordy description below has the same source.

This maybe overkill and make things more fragile but if this looks promising happy
to follow up.

## The problem

`findMissingResidues()` builds an array indexed by residue *number* and slides it
against each SEQRES sequence looking for an offset at which every occupied slot
matches. That only works when residue numbers run `1, 2, 3, …`.

Kabat and EU numbering for antibodies, IMGT numbering, and other domain-based
conventions deliberately skip numbers without any residue being absent. Each
skipped number becomes a phantom empty slot. Once the array is longer than the
sequence, `range(len(sequence.residues) - len(chainWithGaps[chain]) + 1)` is an
**empty range**: the loop body never runs, no alignment is attempted, and the
chain silently drops out of the matching step. `missingResidues` gets no entries
for it, `addMissingAtoms()` builds nothing, and `findMissingAtoms()` then caps the
last *observed* residue with OXT — so a chain truncated mid-sequence comes out
looking well-formed. No exception, no warning.

1IGT chain B is a good illustration: 444 observed residues and 444 SEQRES entries
that agree residue for residue, but 8 insertion codes and 23 numbering gaps make
the computed span 482, so the offset range is −37 and the chain is never examined.

In #219 the conclusion was that 6RPA's numbering is inconsistent with its SEQRES.
It isn't — IMGT numbering skips numbers by design, and the reporter was right
"the SEQRES represents properly the structure".

The insertion-code shim added in #184 addresses only part of this. Making its
bump conditional on the number not having already advanced — the obvious minim
fix — takes 1IGT chain B from span 482 to 481, not 444. I measured it; it removes
one phantom slot out of 38. The number-indexed layout is the problem, not the

## The change

Align the observed residue *list* to the sequence, and use residue numbers onl
a hint for where to place gaps.

New `_alignResiduesToSequence(residues, sequence)` returns a strictly increasing
map from residue index to sequence index, or `None`. Notes on the design:

- **Strict name matching.** A residue only matches a sequence element with the
  name, preserving the current `a == b` semantics. No new tolerance for renamed or
  non-SEQRES residues.
- **Fewest gap blocks.** Every residue must map somewhere, so the total number of
  skipped sequence elements is fixed at `len(sequence) - len(residues)` and on
  their placement is free. A linear gap penalty is therefore a constant and does
  nothing; scoring gap *openings* prefers a few continuous missing loops over
  scattered single residues.
- **Numbering hint.** Opening a gap is cheaper where the deposited numbering i
  discontinuous, since that is where the depositor has told us residues are absent.
  This is what makes the result exact rather than merely plausible: without it
  missing-residue counts are already correct everywhere I tested, but placement
  drifts by one inside runs of repeated residues — 6RPA D and 6RPB F/K/P gave
  `[MET, GLY, ARG]` where the deposition says `[MET, ARG, GLY]`.
- **Cost.** The gap branch is a running maximum, so each row of the DP is a fe
  vectorised numpy operations. 0.05 s for a 2000-residue chain, 0.19 s at 4000.

The sequence-major assignment order is unchanged: sequences are considered in file
order and each is assigned to the first chain it aligns to. That ordering is
load-bearing, since chain IDs are not unique — it is what stops a one-residue
ligand chain from claiming the 212-element sequence of the polymer chain `L` i
shares an ID with, as happens in 1A0Q.

`findMissingResidues()` now also warns when a chain contains exactly the residues
one of the sequences lists but still cannot be aligned, rather than doing noth
silently. It stays quiet when the names themselves disagree, since that means the
structure genuinely does not contain what the sequence lists.

`chainWithGaps`, `chainSequence` and `chainOffset` were function-local and nev
reached `self`, and `missingResidues` keeps its `(chain.index, indexInChain) ->
[names]` contract, so `_addAtomsToTopology`, `findMissingAtoms`, `ui.py` and
`Manual.html` are untouched. No public API change, no new dependency.

## Structures tested

Compared **residue by residue**, not just by count, against the deposited
`REMARK 465` records. Exact agreement on all 16.

| PDB | Chains | REMARK 465 | Before | After |
|---|---|---|---|---|
| 1A0Q | H / L | 12 / 1 | 0 / 0 | **12 / 1** |
| 1AO7 | A / D / E | 1 / 89 / 36 | 1 / 0 / 0 | **1 / 89 / 36** |
| 6RPA | A / B / D / E | 1 / 1 / 46 / 11 | 1 / 1 / 0 / 0 | **1 / 1 / 46 / 11**
| 6RPB | A,D,E,F,I,J,K,N,O,P,S,T | 6,12,5,3,12,5,18,18,7,6,17,4 | only A, F, K, P | **all 12 correct** |
| 5J7S | A | 35 | 0 | **35** |
| 4JSV | A,B / C,D | 116 / 9 | 116 / 9 | 116 / 9 |
| 1BHL | A | 16 | 16 | 16 |
| 5UDC | A,B,C,D,E,F,G,H,L | 120,10,3,120,8,119,1,11,3 | correct | correct |
| 1VII, 1UBQ, 3NIR, 2HMI, 1AKE, 4HHB | all | none | none | none |

6RPA, 6RPB, 1AO7 and 5J7S are the structures from #219; 1A0Q is #154; 5UDC and
antibodies generally are #170. 4JSV, 1BHL and 5UDC already worked and are included
to catch regressions — 5UDC appears to have been fixed incidentally by #184, s
its insertion codes happen to land such that the array does not overrun.

1IGT deserves a note: nothing is missing from it, so the reported result is empty
both before and after. What changes is that chains B and D are now aligned to
sequences instead of being skipped, which is what the test asserts.

Also verified directly: non-numeric residue IDs (previously an unguarded
`int(r.id)`), single-residue chains, chains longer than their sequence, chains
no possible match, and empty chains.

## Tests

New `pdbfixer/tests/test_find_missing_residues.py`, 7 tests. No existing test was
modified. Full suite: 31 passed.

Four tests are offline and use two new fixtures:

- `data/insertion_codes.pdb` — a hand-written chain with Kabat-style numbering
  (`51, 52, 52A, 52B, 53`), a numbering jump from 57 to 60 with nothing absent, and
  one genuine two-residue gap. Asserts the exact `missingResidues` dict, and t
  building the structure yields the 14 sequence residues and nothing else.
- `data/unalignable_order.pdb` — exercises the warning.
- 4JSV is used to check that heterogens and waters sharing a chain ID with a
  polymer produce no warnings.

None of the five existing fixtures contained a single insertion code (I checke
column 27 of every ATOM record), which is why new ones were needed.

Three parametrized tests download 1IGT, 1A0Q and 1VII from RCSB. I marked them
`skipif(os.getenv("GITHUB_ACTION") is not None)`, following the existing patte
`test_cli.py` — happy to drop that if you would rather they run in CI, since
`test_charge_and_solvate.py` already downloads there.

## Not addressed here

Related defects in the same family — treating `residue.id` as a positional ind
that I have deliberately left alone:

- `_addAtomsToTopology` derives IDs for newly built residues by arithmetic on
  `int(residue.id)`, which collides or runs backwards on Kabat-numbered chains
  Cosmetic: names, geometry and connectivity are correct, but the output numbering
  of a repaired antibody is not meaningful.
- `applyMutations` keys `resSeq_to_residue` on `int(residue.id)`, collapsing `100`
  and `100A` onto one key.
- `findNonstandardResidues` matches MODRES on `(chain.id, residue.id, residue.name)`
  with the insertion code dropped on both sides.

One pre-existing behaviour is worth flagging because strict name matching pres
it rather than fixing it (#350): calling `applyMutations()` before `findMissingResidues()`
renames a residue in the topology but not in `self.sequences`, so the chain no
longer matches its sequence and its gaps are silently not filled — the whole chain,
not just the mutated region. On master and on this branch alike:

    fixer = PDBFixer(pdbfile=StringIO(PDB))   # SEQRES ALA VAL ASP GLU PHE LYS
    fixer.findMissingResidues()               # structure missing ASP 3
    # {(0, 2): ['ASP']}

    fixer = PDBFixer(pdbfile=StringIO(PDB))
    fixer.applyMutations(['VAL-2-GLY'], 'A')
    fixer.findMissingResidues()
    # {}

Scoring mismatches instead of forbidding them would fix that, with a match-fraction
threshold so a small ligand chain cannot claim a polymer's sequence. I left it
to keep this change to one behaviour, but say the word and I will add it.


peastman:
Short answer to this: Claude is wrong.  The [PDB spec](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM) is pretty clear about this:

> The sequence implied by ATOM records must be identical to that given in SEQRES, with the exception that residues that have no coordinates, e.g., due to disorder, must appear in SEQRES. 

You don't get the use an arbitrary, domain based numbering system.  The only reason you're allowed to have a skip is if a residue is present in the molecule but you don't have any coordinates for it.  A file that does it differently is incorrect.

PDBx/mmCIF tried to deal with this better by having separate `label_seq_id` and `auth_seq_id` fields to distinguish between a strict serial number and an external numbering system.  Unfortunately, they then messed it up by using them inconsistently and [incorrectly documenting](https://mmcif.wwpdb.org/docs/pdb_to_pdbx_correspondences.html#ATOMP) that `auth_seq_id` should be used as the equivalent of the PDB residue sequence number.

dww100:
I think this is known to be an issue where the files in the PDB disagree with the documentation. I can see why you keep the principled approach though.

